PR #2, opened 2021-01-23
doc: configure dartls according to nvim-lspconfig

muzaffar-omer:
Thank you @akinsho for this awesome plugin!, can't wait to see you releasing it officially.

During setup, I wrongly had the impression this plugin also configures the `dartls`. But after sometime, I realised I must configure `dartls` according to `nvim-lspconfig` to get the LSP things.

I am tempted to mention that `dartls` should be configured according to the instructions in `nvim-lspconfig`. It could save a lot of time for other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ or using `packer.nvim`
 use {"akinsho/flutter-tools.nvim", requires = {"neovim/nvim-lspconfig"}}
 ```
 
-Currently this plugin depends on `nvim-lspconfig` for some default setup this might change.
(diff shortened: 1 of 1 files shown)


akinsho:
@muzaffar-omer this plugin does configure dartls automatically you shouldn't have to set it up yourself. You just need to have nvim-lspconfig loaded before this plugin.

What does your setup look like? Just calling setup should be enough

muzaffar-omer:
> @muzaffar-omer this plugin does configure dartls automatically you shouldn't have to set it up yourself. You just need to have nvim-lspconfig loaded before this plugin.
> 
> What does your setup look like? Just calling setup should be enough

Thank you @akinsho!

You are right, I figured out an issue in my `$PATH`. Somehow, the language server was not visible, even to `nvim-lspconfig`. But for the timebeing it works with `cmd` option in `dartls` config.

I am closing this PR.

PR #4, opened 2021-02-23
feat: add outline guides

akinsho:
Add Widget outline guides, using the recently added `extmark` overlay functionality in `nvim` see https://github.com/neovim/neovim/pull/13952, for details

image

#### TODO
- [x] Fix offset for child nodes occasionally being incorrect
- [x] Make highlight groups configurable
- [x] Handle navigating between buffers
- [ ] Stretch: add tests


akinsho:
This is not perfect yet but less buggy and can be iterated on in future PRs. It's currently off by default till stable

PR #21, opened 2021-04-02
Remove dependency on lspconfig

akinsho:
lspconfig is a thin wrapper around starting an lsp server which in actuality isn't that tough to do without it. So this PR removes the dependency since it is not absolutely necessary. Whilst very likely users will have it installed it isn't guaranteed. Also this seems like it gives me more direct control over the lsp server so functionality or configurations won't change underfoot.

EDIT: also tbh I wanted to get more familiar with the native lsp functionality rather than work with/around an abstraction layer

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ or using `packer.nvim`
 use "akinsho/flutter-tools.nvim"
 ```
 
+NOTE: flutter tools does not depend on `nvim-lspconfig`. The two can co-exist but please ensure
+you do **NOT** configure `dartls` using `lspconfig`. It will be automatically set up by this
+plugin instead.
+
 To set it up
 
 ```lua
--- a/README.md
+++ b/README.md
@@ -23,17 +23,15 @@ to allow users to easily develop flutter apps using neovim.
 using `vim-plug`
 
 ```vim
-Plug "neovim/nvim-lspconfig"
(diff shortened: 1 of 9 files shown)


akinsho:
@nghialm269 you've been quite helpful reporting issues in the past. I'm working on removing the dependency of lspconfig. Would you be able to test out this PR and see if it works for you? Seems fine on my end and I'm planning on merging it shortly just wanted a sense check

nghialm269:
@akinsho I just tested this branch now and it works fine on my end, too.

btw, do you happen to get the extract local variable, extract widget, ... code actions work?

akinsho:
@nghialm269 are they no longer working on this branch for you. I might be missing some parameters that were being handled by lspconfig I'll check locally

nghialm269:
@akinsho no I never get them working even with lspconfig alone, I was just wondering if it's just me or not.

akinsho:
@nghialm269 I can get most code actions to work but I think I've had some issues with both of those as well. One thing to note is that the native lsp currently doesn't do a good job of telling you if a code action fails for a valid reason since I don't think it currently prints the explanation. It could also be that it needs to report some different capabilities but no one has added them yet.

I'll investigate a little bit more as a separate issue, not 100% sure about those particular actions, can't remember if they work for me or not but a quick test didn't work just now.

PR #24, opened 2021-04-03
outline: Fix start_line/col position

simrat39:
for whatever reason both of these are one less than where they should be, this fixes the cursor being one line above than the actual class/widget when navigating through the outline


simrat39:
Before:
f_before

After:
f_after

akinsho:
@simrat39 thanks for the PR. I've just tried it out locally and it seems like it actually regresses the behaviour but depends on the problem you were actually trying to solve.

The current behaviour below:
outline_current

If you watch the GIF you see that once you place your cursor on a symbol it gets highlighted in the outline.

Behaviour in your PR:
outline_branch
In your PR ranges are all moved down one so some symbols which are single line ones never get highlighted. And placing your cursor on a symbol doesn't highlight it till you move one line down.

simrat39:
@akinsho Should be fixed with this though I wonder if there is a better way to handle this

akinsho:
@simrat39 that looks like a more conservative fix which I'm in favour of except I don't really understand why that change is needed which worries me a little about if it will work for all cases. I think the lsp returns both a `range` and a `codeRange` and maybe the `codeRange` is what should be used here.

See [here](https://github.com/dart-lang/sdk/tree/master/pkg/analysis_server/tool/lsp_spec#darttextdocumentpublishflutteroutline-notification) for details.

simrat39:
> @simrat39 that looks like a more conservative fix which I'm in favour of except I don't really understand why that change is needed which worries me a little about if it will work for all cases. I think the lsp returns both a `range` and a `codeRange` and maybe the `codeRange` is what should be used here.
> 
> See [here](https://github.com/dart-lang/sdk/tree/master/pkg/analysis_server/tool/lsp_spec#darttextdocumentpublishflutteroutline-notification) for details.

Looks like the outline code actually uses codeRange 
https://github.com/akinsho/flutter-tools.nvim/blob/main/lua/flutter-tools/outline.lua#L167

Anyways there's element.range in every node but it seems like both codeRange and element.range give the same line and the column is even further behind (or sometimes the same) in element.range

```
  print(vim.inspect({
      codeRange_start_line = node.codeRange.start.line,
      codeRange_start_col = node.codeRange.start.character,
      element_range_start_line = node.element.range.start.line,
      element_range_start_col = node.element.range.start.character
  }))

```
Gives:
```
{messages                                                                                                                                                                   
  codeRange_start_col = 50,                                                                                                                                                 
  codeRange_start_line = 11,                                                                                                                                                
  element_range_start_col = 56,                                                                                                                                             
  element_range_start_line = 11                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 2,                                                                                                                                                  
  codeRange_start_line = 13,                                                                                                                                                
  element_range_start_col = 13,                                                                                                                                             
  element_range_start_line = 13                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 0,                                                                                                                                                  
  codeRange_start_line = 22,                                                                                                                                                
  element_range_start_col = 5,                                                                                                                                              
  element_range_start_line = 22                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 9,                                                                                                                                                  
  codeRange_start_line = 24,                                                                                                                                                
  element_range_start_col = 9,                                                                                                                                              
  element_range_start_line = 24                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 30,                                                                                                                                                 
  codeRange_start_line = 24,                                                                                                                                                
  element_range_start_col = 30,                                                                                                                                             
  element_range_start_line = 24                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 0,                                                                                                                                                  
  codeRange_start_line = 27,                                                                                                                                                
  element_range_start_col = 6,                                                                                                                                              
  element_range_start_line = 27                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 2,                                                                                                                                                  
  codeRange_start_line = 29,                                                                                                                                                
  element_range_start_col = 19,                                                                                                                                             
  element_range_start_line = 29                                                                                                                                             
}                                     
```

simrat39:
> @simrat39 that looks like a more conservative fix which I'm in favour of except I don't really understand why that change is needed which worries me a little about if it will work for all cases. I think the lsp returns both a `range` and a `codeRange` and maybe the `codeRange` is what should be used here.
> 
> See [here](https://github.com/dart-lang/sdk/tree/master/pkg/analysis_server/tool/lsp_spec#darttextdocumentpublishflutteroutline-notification) for details.

Looks like the Outline code already uses codeRange: https://github.com/akinsho/flutter-tools.nvim/blob/main/lua/flutter-tools/outline.lua#L167

Anyways the lsp returns element.range for every node, but it seems like the line it gives is the same and the column is a bit ahead or sometimes the same

```
  print(vim.inspect({
      codeRange_start_line = node.codeRange.start.line,
      codeRange_start_col = node.codeRange.start.character,
      element_range_start_line = node.element.range.start.line,
      element_range_start_col = node.element.range.start.character
  }))
```
Gives: ```

{                                                                                                                                                                           
  codeRange_start_col = 50,                                                                                                                                                 
  codeRange_start_line = 11,                                                                                                                                                
  element_range_start_col = 56,                                                                                                                                             
  element_range_start_line = 11                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 2,                                                                                                                                                  
  codeRange_start_line = 13,                                                                                                                                                
  element_range_start_col = 13,                                                                                                                                             
  element_range_start_line = 13                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 0,                                                                                                                                                  
  codeRange_start_line = 22,                                                                                                                                                
  element_range_start_col = 5,                                                                                                                                              
  element_range_start_line = 22                                                                                                                                             
}                                                                                                                                                                           
{                                                                                                                                                                           
  codeRange_start_col = 9,                                                                                                                                                  
  codeRange_start_line = 24,                                                                                                                                                
  element_range_start_col = 9,                                                                                                                                              
  element_range_start_line = 24                                                                                                                                             
}  
...
...               
```

akinsho:
@simrat39 I haven't forgotten about this by the way just doing a lot of different things, also I haven't quite figured out what the right solution is here yet. From what you say it seems like the magic number of 2 for the column isn't going to work all the time and will sometimes put the user in the wrong position.

I'm sure there must be the right coordinates in the lsp response and it's just a matter of converting that in between lua and nvim's indexing

simrat39:
> @simrat39 I haven't forgotten about this by the way just doing a lot of different things, also I haven't quite figured out what the right solution is here yet. From what you say it seems like the magic number of 2 for the column isn't going to work all the time and will sometimes put the user in the wrong position.
> 
> I'm sure there must be the right coordinates in the lsp response and it's just a matter of converting that in between lua and nvim's indexing

i'm not sure but maybe we can find something here: https://github.com/Dart-Code/Dart-Code/blob/master/src/extension/flutter/flutter_outline_view.ts#L357-L377
ill take a look at it later

akinsho:
Going to close this out as I merged something similar to it in 8f301cd, I believe the issue was that I was minus-ing one from the start col in the parsing function but tbh I'm still not clear on why adding another one is required to have things start at the right column. I wonder if it might be a 1-based index API thing 🤔. Either way, this seems to work for now. I've also added you as a co-author of the commit as it was due to your work in this PR, but it's quite out of date and hard to incorporate as is.

PR #31, opened 2021-04-26
Feature: Add outline CodeActions support

simrat39:
* VsCode allows wrapping widgets with center and a couple of others by
   right clicking them in the outline, so lets take it a notch further
   and add full code actions support

This works and all but is pretty wip regarding code style and some other stuff, like forcing a refresh after the action


simrat39:
code_actions

akinsho:
@simrat39 this looks really cool. I wonder if the handler could be something a user could configure. For example I use lsp saga for code actions and it has a code action function which maybe I would be able to pass to this and if the user doesn't specify one then this just uses this default rendering? or maybe it's more that lspsaga should be overriding the lsp handler for all code actions in the first place 🤔

akinsho:
Regarding code style I've added an action to automatically format main when a change is pushed so hopefully it will pick up PRs and automatically format them.

If you wanted to do it on your end I use [`stylua`](https://github.com/JohnnyMorganz/StyLua) to format the project.

simrat39:
@akinsho I'll use stylua, thanks. As for causing a refresh, I have tested a bit and found out that it's pretty much impossible unless we have weird hacks or we rewrite the entire code actions handler.

simrat39:
Above mentioned weird hacks: 

```
            api.nvim_buf_attach(ca_buf, false, {on_detach=function (_,_)
                vim.fn.win_gotoid(code_wins[1])
                vim.fn.timer_start(500, function ()
                    vim.fn.win_gotoid(outline_wins[1])
                end)
            end})
```

akinsho:
@simrat39 when you say it doesn't refresh do you mean that the outline window doesn't show the newly added outline? I wonder if you can request a new outline for a buffer rather than just listening for one? I also wonder what `Dart-Code` do? or emacs-dart-lsp I sometimes look at their code for inspiration

simrat39:
@akinsho Yes that's what I mean by making it refresh. The thing is that we don't exactly know when the codeaction will be applied, unless we write our own handler, because if we use the default handler it does not wait for the user input and the code keeps evaluating, which makes it refresh before any changes are made. That's why the hack is there, it attaches itself to the code actions buffer, and when that buffer is deleted it waits a bit and refreshes the outline. But a drawback that has is that not all code action handlers create a floating window, LspSaga and nvim-lsputils do but the default one included in neovim shows a prompt so this approach won't work there as well.

As for how dart-code does it, I don't think they show code actions at all, but only have a few set choices, so they know when the choice will be applied and can trigger a refresh after. 

I think the most optimal solution would be to write our own handler but that's pretty bad for if we want to keep things uniform

akinsho:
@simrat39 thanks for clarifying hmm seems some sort of refresh is required then I agree about not creating custom handlers or anything. Regarding the hack could this be done slightly differently by doing something like
```lua
  api.nvim_buf_attach(ca_buf, false, {on_detach=function (_,_)
     vim.api.nvim_set_current_win(code_wins[1])
       vim.defer_fn(function () -- <-- use defer as it's all lua and I think timer leaves things hanging around unless stopped maybe
        vim.fn.win_gotoid(outline_wins[1])
         api.nvim_buf_detach(ca_buf) -- <- clean up since I'd rather this just didn't recreate loads of listeners
      end, 500)
 end})
```

Tbh I really don't think it not refreshing is that big a deal it's a nice to have feature *at best* I think we could just place the cursor in the code window and that would trigger a refresh at some point. It's nice to be able to add widgets from the outline window but it isn't a replacement from editing your code in the buffer so I don't think we need to go too far to make that possible. 

In either case if the tweaked hack will work for most cases, without causing side effects elsewhere then I'm fine with that and it can be iterated on in the future

tjdevries:
Drive by comment: telescope already has code action support. You may be able to use that as the interface for selecting one without having to do much work.

akinsho:
Final version, re-uses this plugin's float UI and implements the hack from above for now. Will investigate a better solution in the future
code_action_flutter

Issue #9, opened 2021-02-24
Improve/Document DevTools integration


akinsho:
I've done some preliminary work to integrate dev tools. This should be surfaced and documented and also tweaked to make it easier to use

akinsho:
Gonna close this out since it's kind of ambiguous in scope and some helpful improvements have already resulted from it. If someone wants to pitch a specific improvement for the dev tools please open a new PR

PR #41, opened 2021-05-11
Add command to copy profiler url

Charly6596:
This PR adds a command to copy the profiler url to the clipboard.
In order to execute the command the app and our own DevTools server must be running (FlutterRun + FlutterCopyProfilerUrl commands)

- [X] Command implemented
- [X] Command registered in the `setup_commands` function
- [X] Command added to the telescope menu
- [x] Documentation updated

### Concerns
I've found the following issues while working on this
- As mentioned here, #9 there are some issues with how the DevTools are managed: We don't stop the process. Would be great to have a way to reuse an existing dart devtools process if it's already running and gather its info.
- Chrome and Android have different console outputs regarding the profiler url as mentioned in the [`search_profiler_url`](https://github.com/akinsho/flutter-tools.nvim/compare/main...Charly6596:copy-devtools-link?expand=1#diff-96c4b32009844045b7f1ce3515556a45d330b4b0c2567fa9fed85d0efb0a782fR27) comments. The Android execution runs another dart tools process and outputs an already formatted link to the profiler. The `flutter run` command has a flag to specify the dev tools url, `--devtools-server-address`. We could (optionally) start our own devtools server before starting the app and use it to have more control over it.
- Commands that shouldn't be executed are shown in the menu without any distinction. I'm talking about reloading, restarting the app while the app is not running for example. In `coc-flutter` they hide those commands, wonder if something similar could be done, something like marking the commands as not executable

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ NOTE: To configure the highlight colour you can override the `FlutterWidgetGuide
 - `FlutterOutline` - Opens an outline window showing the widget tree for the given file.
 - `FlutterDevTools` - Starts a Dart Dev Tools server.
 - `FlutterCopyProfilerUrl` - Copies the profiler url to your system clipboard (+ register). Note that commands `FlutterRun` and
-`FlutterDevTolls` must be executed first.
+`FlutterDevTools` must be executed first.
 
 ## Debugging
 
--- a/README.md
+++ b/README.md
@@ -137,10 +137,13 @@ NOTE: To configure the highlight colour you can override the `FlutterWidgetGuide
 - `FlutterRun` - Run the current project. This needs to be run from within a flutter project.
 - `FlutterDevices` - Brings up a list of connected devices to select from.
 - `FlutterEmulators` - Similar to devices but shows a list of emulators to choose from.
-- `FlutterReload` - Reload the running project
-- `FlutterRestart` - Restart the current project
-- `FlutterQuit` - Ends a running session
-- `FlutterOutline` - Opens an outline window showing the widget tree for the given file
+- `FlutterReload` - Reload the running project.
+- `FlutterRestart` - Restart the current project.
+- `FlutterQuit` - Ends a running session.
+- `FlutterOutline` - Opens an outline window showing the widget tree for the given file.
+- `FlutterDevTools` - Starts a Dart Dev Tools server.
+- `FlutterCopyProfilerUrl` - Copies the profiler url to your system clipboard (+ register). Note that commands `FlutterRun` and
+`FlutterDevTolls` must be executed first.
 
 ## Debugging
 
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -5,6 +5,7 @@ local devices = require("flutter-tools.devices")
 local config = require("flutter-tools.config")
 local executable = require("flutter-tools.executable")
 local dev_log = require("flutter-tools.log")
+local dev_tools = require("flutter-tools.dev_tools")
 
 local api = vim.api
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -137,10 +137,13 @@ NOTE: To configure the highlight colour you can override the `FlutterWidgetGuide
...
-- `FlutterRestart` - Restart the current project
-- `FlutterQuit` - Ends a running session
-- `FlutterOutline` - Opens an outline window showing the widget tree for the given file
+- `FlutterReload` - Reload the running project.
+- `FlutterRestart` - Restart the current project.
+- `FlutterQuit` - Ends a running session.
+- `FlutterOutline` - Opens an outline window showing the widget tree for the given file.
+- `FlutterDevTools` - Starts a Dart Dev Tools server.
+- `FlutterCopyProfilerUrl` - Copies the profiler url to your system clipboard (+ register). Note that commands `FlutterRun` and
+`FlutterDevTolls` must be executed first.
```
akinsho, line 146:
```suggestion
`FlutterDevTools` must be executed first.
```

lua/flutter-tools/commands.lua
```diff
@@ -13,6 +14,34 @@ local M = {}
...
+local function search_profiler_url(data)
+  -- We already have it, stop checking messages
+  if profiler_url then
+    return
+  end
+
+  -- Chrome
+  -- Debug service listening on ws://127.0.0.1:44293/heXbxLM_lhM=/ws
+  local m = data:match("Debug service listening on (ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
+  if(not m) then
```
akinsho, line 30: Parens aren't actually necessary in `lua` conditionals, tbh there is a CI formatting workflow that will run after this is merged anyway so slightly moot point but just FYI

akinsho (commented):
@Charly6596 this LGTM 👍🏾 will give it a try later on today after work but happy to merge it otherwise.

akinsho:
@Charly6596 depending on if you wanted to tackle the issues you raised here (I'm fine for these to be in another PR 🤷🏾 )

1. Hanging devtools job - the job can be stored in a module variable and then when launched you can create a `VimLeavPre` autocommand that calls a module function to stop the dev tools job.
3. Commands showing up out of context - in `menu.lua` when constructing the list of items for telescope you could create a helper function to be exported from the `commands` file which is to just check if the `run` job is still alive and then check it in `menu.lua` if it is still running the you can hide the no longer applicable options etc.

lua/flutter-tools/commands.lua
Charly6596: Thanks, I'm new to `lua` so this kind of feedback is always helpful :+1:

Charly6596:
> @Charly6596 depending on if you wanted to tackle the issues you raised here (I'm fine for these to be in another PR 🤷🏾 )
> 
> 1. Hanging devtools job - the job can be stored in a module variable and then when launched you can create a `VimLeavPre` autocommand that calls a module function to stop the dev tools job.
> 2. Commands showing up out of context - in `menu.lua` when constructing the list of items for telescope you could create a helper function to be exported from the `commands` file which is to just check if the `run` job is still alive and then check it in `menu.lua` if it is still running the you can hide the no longer applicable options etc.

Sure, I think I could do that, thanks for the suggestions. Would we need to remove the commands from vim too, so we couldn't execute them even if we typed it?
image

akinsho:
@Charly6596 I think not re. removing the vim commands, dynamically adding and clearing vim commands is a lot messier and more unwieldy than controlling telescope menu items. I wouldn't do it in this PR in either case because that will require a bit of finessing (and tbh I don't think it will be worth it to dynamically changed registered vim commands)

PR #46, opened 2021-05-15
feat: Add format instead of only prefix for labels

tjdevries:
I wanted to be able to have closing labels do something like `</ $NAME >` (or at least play with the idea).

This change should be backwards compat with `prefix`

### DIFF
--- a/lua/flutter-tools/labels.lua
+++ b/lua/flutter-tools/labels.lua
@@ -9,16 +9,25 @@ local function render_labels(labels, opts)
   opts = opts or {}
   local highlight = opts.highlight or "Comment"
 
+  -- Have ot use `rawget` to override what the default format is that we provide in config.
+  local user_format = rawget(opts, "format")
+
   local prefix = opts.prefix
-  local format = opts.format
-  if prefix and format then
+  if prefix and user_format then
     error(
       "[flutter-tools.labels] Cannot have both prefix and format specified" .. vim.inspect(opts)
     )
   end
 
-  if not format then
+  -- A bit complicated, but basically just keeps backwards compat of passing "prefix"
+  -- while still allowing users to pass "format" instead.
+  local format
+  if user_format then
+    format = user_format
+  elseif prefix then
     format = prefix .. "%s"
+  else
+    format = opts.format
   end
 
   for _, item in ipairs(labels) do
--- a/lua/flutter-tools/labels.lua
+++ b/lua/flutter-tools/labels.lua
@@ -7,13 +7,24 @@ local namespace = api.nvim_create_namespace("flutter_tools_closing_labels")
 local function render_labels(labels, opts)
   api.nvim_buf_clear_namespace(0, namespace, 0, -1)
   opts = opts or {}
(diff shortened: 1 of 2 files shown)


tjdevries:
Although now I'm kind of thinking, what if you allowed a function to be passed that returned the virtual text chunk (so it could be text + highlights if you wanted to configure that. it would be basically just as easy to implement)

lua/flutter-tools/labels.lua
```diff
@@ -7,13 +7,33 @@ local namespace = api.nvim_create_namespace("flutter_tools_closing_labels")
...
   opts = opts or {}
-  local highlight = opts and opts.highlight or "Comment"
-  local prefix = opts and opts.prefix or "// "
+  local highlight = opts.highlight or "Comment"
+
+  -- Have ot use `rawget` to override what the default format is that we provide in config.
+  local user_format = rawget(opts, "format")
+
+  local prefix = opts.prefix
+  if prefix and user_format then
```
akinsho, line 16: The `config` module actually already has a mechanism for deprecations, you just need to specify what the fallback key is and the message to show the user
https://github.com/akinsho/flutter-tools.nvim/blob/a2a67dae9ff43d95caddf4d68cf5afcd4cb9d65d/lua/flutter-tools/config.lua#L61

lua/flutter-tools/labels.lua
```diff
@@ -7,13 +7,33 @@ local namespace = api.nvim_create_namespace("flutter_tools_closing_labels")
...
+    error(
+      "[flutter-tools.labels] Cannot have both prefix and format specified" .. vim.inspect(opts)
+    )
+  end
+
+  -- A bit complicated, but basically just keeps backwards compat of passing "prefix"
```
akinsho, line 22: I'd actually rather break this (mean as that sounds) whilst this plugin is still fairly small/doesn't have that many users so I can keep things as simple as possible. I think if the user has overridden `prefix` with the deprecation table it should warn them of the change and they can change their config

akinsho (commented):
@tjdevries thanks for the PR 🙌🏾 , I made a few comments, nothing major. Think we can get away with re-using some existing logic and doing a hard switch here vs. trying to maintain backward compat since this plugin likely doesn't have very many users atm.

Regarding the function option I agree it would be easy enough to add but I'd rather wait to see if enough people end up wanting that, it appears on almost every line so I don't think it needs to be very fancy.

tjdevries:
Well the thing I was thinking for the function was I kind of wanted to be able to do different highlights for the ` // ` at the beginning of the line and then something else for the actual closing label. So doing a function might just be easier to make that, rather than having some convluted mess of options.

akinsho:
That's fair enough, in which case I'd rather approach this by leaving the existing option as is, and offering a second option which is a function that does what you describe.

I'd approach it this way because *by far* the majority of issues I've seen are from users who are just trying to get to grips with nvim/lua in the first place so I'd definitely rather there be a very simple option much like coc or vscode where all they have to do is change a string and a more advanced option using a lua function for people who want to do fancier things.

PR #50, opened 2021-05-16
Hide disabled commands in telescope menu

Charly6596:
~~~This depends on #44. The way I detect whether the devtools server is running will be changed once that's merged~~

- [x] ~~Documentation~~
- [X] Hide disabled commands in telescope menu
- [x] Fix conflicts with #44 

There are 2 main conditions to group commands: **devtools server running** and **flutter run running**
### Always enabled
- `FlutterDevices`
- `FlutterEmulators`
- `FlutterOpenOutline`
- `FlutterLogClear`

### Devtools server
#### Running
- `FlutterCopyProfileUrl`
#### Not running
- `FlutterDevTools`

### Flutter run
#### Running
- `FlutterReload`
- `FlutterRestart`
- `FlutterVisualDebug`
- `FlutterQuit`

#### Not running
- `FlutterRun`

### Devtools not running, flutter not running
![image](https://user-images.githubusercontent.com/13708100/118396473-fc824700-b64f-11eb-8580-60227abe4b9f.png)

### Devtools not running, flutter running
![image](https://user-images.githubusercontent.com/13708100/118397210-40c31680-b653-11eb-8b5c-d6bb95b4d5c9.png)

### Devtools running, flutter not running
![image](https://user-images.githubusercontent.com/13708100/118396538-3e12f200-b650-11eb-8bd1-a298fcd14b3c.png)

### Devtools running, flutter running
![image](https://user-images.githubusercontent.com/13708100/118396612-8fbb7c80-b650-11eb-86fb-80fcf1283fa3.png)

### DIFF
--- a/lua/flutter-tools/dev_tools.lua
+++ b/lua/flutter-tools/dev_tools.lua
@@ -94,4 +94,9 @@ function M.get_url()
   return devtools_url
 end
 
+---@return boolean
+function M.is_running()
+  return devtools_url ~= nil
+end
+
 return M
--- a/lua/flutter-tools/dev_tools.lua
+++ b/lua/flutter-tools/dev_tools.lua
@@ -180,4 +180,9 @@ function M.on_flutter_shutdown()
 
 end
 
+---@return boolean
+function M.is_running()
+  return devtools_profiler_url ~= nil and devtools_url ~= nil
+end
+
 return M
--- a/lua/flutter-tools/dev_tools.lua
+++ b/lua/flutter-tools/dev_tools.lua
@@ -160,7 +160,8 @@ end
 
 ---@return boolean
 function M.is_running()
-  return devtools_profiler_url ~= nil or devtools_url ~= nil 
+  return devtools_profiler_url ~= nil or devtools_url ~= nil
+end
 
 ---@return string? devtools_profiler_url the url including the devtools url and the app url. Follows the format `devtools_url/?uri=app_url`
 ---@return boolean? server_running true if there is a `devtools_url` available but couldn't build the url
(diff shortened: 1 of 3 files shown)


Charly6596:
oh forgot to ask, @akinsho how do you think we should add this to the documentation, I can't think of a title that fits this
I would basically paste the always enabled / devtools / flutter thing with a title

akinsho:
@Charly6596 I'm not sure it needs to be explicitly stated tbh, In my mind this is is just logical behaviour rather than something to explain in the README i.e. some things make sense to be available in some contexts and somethings don't hopefully users will just implicitly benefit from it without thinking.

Charly6596:
@akinsho right, makes sense, Im done with this one then

lua/flutter-tools/dev_tools.lua
```diff
@@ -157,6 +157,11 @@ function M.get_url()
   return devtools_url
 end
 
+
+---@return boolean
+function M.is_running()
+  return devtools_profiler_url ~= nil or devtools_url ~= nil
```
akinsho, line 163: @Charly6596 this is a syntax error I think from a bad merge

akinsho (changes requested):
@Charly6596 there seems to be a missing end in the `M.is_running()`function in `dev_tools`

lua/flutter-tools/dev_tools.lua
Charly6596: It should be fixed now

akinsho (approved):
Awesome thanks for this @Charly6596 🚀 🙌🏾

PR #52, opened 2021-05-25
autostart devtools if not detected

### LINKED ISSUE #48: Autostart devtools
Currently, if the plugin doesn't detect the complete devtools url from the `flutter run` console output (the one including devtools server and app url) we would need to start the devtools server manually. 
Would be nice to have a configurable option to autostart the server if the plugin doesn't detect it.
This happens when running web.

For more context, see https://github.com/akinsho/flutter-tools.nvim/pull/44#issuecomment-841721090 and https://github.com/akinsho/flutter-tools.nvim/issues/44#issuecomment-841730696

Charly6596:
closes #48 
also fixed a bug in which the value of `profiler_url` is overriden if devtools is not started

- [x] Update documentation
- [X] Implement feature
- [X] Works on my machine (enabling/disabling the config works as expected)

A new configuration has been added. This feature is disabled by default
```
dev_tools = {
  autostart = false
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ require("flutter-tools").setup {
   dev_log = {
     open_cmd = "tabedit", -- command to use to open the log buffer
   },
+  dev_tools = {
+    autostart = false, -- autostart devtools server if not detected
+  },
   outline = {
     open_cmd = "30vnew", -- command to use to open the outline buffer
   },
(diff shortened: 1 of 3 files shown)


akinsho:
Thanks again for the PR @Charly6596 🥇

PR #58, opened 2021-06-09
fix: don't analyse flutter sdk packages

### LINKED ISSUE #57: Lagging when navigating Flutter SDK
This may not be related to `flutter-tools` but instead https://github.com/hrsh7th/nvim-compe but would love to hear if you have a solution for it.

When navigating the Flutter source code, it starts lagging immensely which I think is related to how there are errors everywhere and this also hinders navigating deeper in the SDK, for example navigating to `StatefulWidget` from here.
image 

Do you have this issue? Or do you have any specific configurations done to not get errors when navigating the Flutter SDK?

akinsho:
@RobertBrunhage can you try this PR and see if it still fixes the lag mention in #57. This adds the exclusion by default, and notes that in the README if a user wanted this behaviour for some reason.

fixes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,8 @@ require("flutter-tools").setup {
     end,
     settings = {
       showTodos = true,
-      completeFunctionCalls = true -- NOTE: this is WIP and doesn't work currently
+      completeFunctionCalls = true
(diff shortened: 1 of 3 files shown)


RobertBrunhage:
Sure thing, using Vim-Plug to install, how would I install this specific branch?

PR #59, opened 2021-06-13
Open devtools in the browser automatically

### LINKED ISSUE #49: Open devtools in the browser automatically
- Open devtools in the browser automatically whenever the server is detected (we either have `devtools_url and profiler_url` or `devtools_profiler_url` available)
- This should be configurable, add an option to opt-out

See https://github.com/akinsho/flutter-tools.nvim/pull/44#issuecomment-841721090 for more context.

Charly6596:
closes #49 
- Feature **enabled** by default
- Fixed a bug in which a devtools server was started even when not needed
- Tested Chrome and Android
- Updated documentation


lua/flutter-tools/dev_tools.lua
```diff
@@ -31,15 +31,31 @@ local function try_get_tools_flutter(data)
   return data:match("Flutter DevTools, a Flutter debugger and profiler, on .+ is available at:%s(https?://127%.0%.0%.1:%d+%?uri=.+)$")
```
Charly6596, line 31: I'm assuming this works on android and ios. Not sure if there is any case in which this message doesn't appear (and flutter doesnt start a devtools server)

lua/flutter-tools/dev_tools.lua
```diff
@@ -31,15 +31,31 @@ local function try_get_tools_flutter(data)
...
 end
 
---- Debug service listening on ws://127.0.0.1:44293/heXbxLM_lhM=/ws
 --- An Observatory debugger and profiler on sdk gphone x86 arm is available at: http://127.0.0.1:46051/NvCev-HjyX4=/
 ---@param data string
 ---@return string?
 local function try_get_profiler_url(data)
   return data:match("An Observatory debugger and profiler on .+ is available at:%s(https?://127%.0%.0%.1:%d+/.+/)$")
```
Charly6596, line 38: Related to my last comment, we need to test whether this match is neccesary (right now is not being used)

lua/flutter-tools/dev_tools.lua
```diff
@@ -31,15 +31,31 @@ local function try_get_tools_flutter(data)
...
+    return data:match("Debug service listening on (ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
+end
+
+local function start_browser()
+
+  local autoopen_browser = require("flutter-tools.config").get("dev_tools").autoopen_browser
+  if not autoopen_browser then return end
+  local url = M.get_profiler_url()
+  if url then
+    vim.fn.system(string.format("xdg-open '%s'", url))
```
akinsho, line 54: A minor improvement you can make here is to use `jobstart` instead of `system` (I know I mentioned system but almost immediately thought of jobstart after but forgot to mention).

The reason being that `system` is a blocking call so if the browser needs to be loaded or the computer is slow it will block momentarily, tbh it's really not a big deal but
using jobstart will allow nvim to keep going and the browser will open when it's ready
e.g.

```lua
    fn.jobstart({<open-command>, <url>}, {detach = true})
```

lua/flutter-tools/dev_tools.lua
```diff
@@ -31,15 +31,31 @@ local function try_get_tools_flutter(data)
...
+--- Debug service listening on ws://127.0.0.1:44293/heXbxLM_lhM=/ws
+-----@param data string
+---@return string?
+local function try_get_profiler_url_chrome(data)
+    return data:match("Debug service listening on (ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
+end
+
+local function start_browser()
+
+  local autoopen_browser = require("flutter-tools.config").get("dev_tools").autoopen_browser
```
akinsho, line 50: nit: I think this variable should be `auto_open` super minor nit, just that two `oo` together makes the word harder to read imo

lua/flutter-tools/dev_tools.lua
```diff
@@ -47,11 +47,11 @@ end
...
 local function start_browser()
 
-  local autoopen_browser = require("flutter-tools.config").get("dev_tools").autoopen_browser
-  if not autoopen_browser then return end
+  local auto_open_browser = require("flutter-tools.config").get("dev_tools").auto_open_browser
+  if not auto_open_browser then return end
   local url = M.get_profiler_url()
   if url then
-    vim.fn.system(string.format("xdg-open '%s'", url))
+    vim.fn.jobstart({"xdg-open", url}, {detach = true})
```
akinsho, line 54: I'm not sure `xdg-open` actually works on `macOS` I use one for work and tend to have a variable in my config that uses `open` for mac and `xdg-open` for linux. That might not be necessary tbh I don't even remember when I added it, but it's worth verifying IMO

akinsho:
@Charly6596 will do a proper review soon, just a bit all over the place atm with a few different things.

akinsho:
@Charly6596 just tried this locally on my Linux machine and the dev tools don't get opened 🤔 . I also noticed that I think the matching will pick up the incorrect dev tools URL. In my logs, I see two messages relating to dev tools printed.

The first one
```
An Observatory debugger and profiler on sdk gphone x86 is available at: http://127.0.0.1:38037/jAe4NZsNtUs=/
```
would be matched but rather than opening the fool fledged dev tools opens not sure what the official name for the tool or page is, but it's not the dev tools.

```
The Flutter DevTools debugger and profiler on sdk gphone x86 is available at: http://127.0.0.1:9100?uri=http%3A%2F%2F127.0.0.1%3A38037%2FjAe4NZsNtUs%3D%2F
```
This one is not matched I think, but opens the dev tools.

I'm on flutter `2.2.1` if that helps/changes things.

PS: sorry about the merge conflict

Charly6596:
@akinsho _It works on my machine_.
Jokes aside, thank you, I'll take a look most likely next week. 
The second link is the one that should match. It opens locally on the port 9100 by default.
No worries about the conflict :smile:

akinsho:
@Charly6596, I'll use the branch during the day tomorrow on my mac and see if I get a result. I tried it on my ubuntu machine with firefox as my default browser and didn't see a new tab open. It also didn't show the notification that the url was copied to my clipboard. Are you also on flutter `2.2.1`?

Charly6596:
> @Charly6596, I'll use the branch during the day tomorrow on my mac and see if I get a result. I tried it on my ubuntu machine with firefox as my default browser and didn't see a new tab open. It also didn't show the notification that the url was copied to my clipboard. Are you also on flutter `2.2.1`?

I'm on 2.0.6

akinsho:
I'm just wondering if the output has changed since looking at the matches in this PR they don't seem to match exactly what is in the log output `The Flutter DevTools debugger and profiler ` line is not really the same as any of the others being matched against

Charly6596:
> I'm just wondering if the output has changed since looking at the matches in this PR they don't seem to match exactly what is in the log output `The Flutter DevTools debugger and profiler ` line is not really the same as any of the others being matched against

Yes, you are right, looks like it changed. Probably we could use a more general expression to match both the old and the new version

Charly6596:
Alright, I've updated and indeed they changed the message. I've changed the match to just look for the url (removing any other text). This would prevent future changes and we should be fine assuming that if we find an url with that format it's the good one (unless they change the url output)

I will write my findings here just in case we need to check in the future.

### Running on chrome:
```
This app is linked to the debug service: ws://127.0.0.1:45351/pSvpl3bUyx8=/ws
Debug service listening on ws://127.0.0.1:45351/pSvpl3bUyx8=/ws
```
Looks like they added a new line (the first one) but kept the old one (second). 
There's a significant difference between this url and the android one, that allows us to identify that we need to run devtools ourselves (it doesn't autostart when running on chrome)

OLD:
```lua
data:match("Debug service listening on (ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
```
NEW:
```lua
data:match("(ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
```

### Running on Android:
The important url here is the one inlcluding the devtools url, because it gives us the port it's running on.
```
An Observatory debugger and profiler on sdk gphone x86 arm is available at: http://127.0.0.1:33103/k_5ewJJ_cE4=/
The Flutter DevTools debugger and profiler on sdk gphone x86 arm is available at: http://127.0.0.1:9100?uri=http%3A%2F%2F127.0.0.1%3A33103%2Fk_5ewJJ_cE4%3D%2F
```
We can delete the function matching "An observatory..." and just match the full url as mentioned earlier

OLD:
```lua
data:match("Flutter DevTools, a Flutter debugger and profiler, on .+ is available at:%s(https?://127%.0%.0%.1:%d+%?uri=.+)$")

```
NEW:
```lua
data:match("(https?://127%.0%.0%.1:%d+%?uri=.+)$")
```

Charly6596:
I've checked the new match expressions and works in both chrome and android. It should also work on the old Flutter version :+1: 
@akinsho  could you confirm whether `xdg-open` works on mac?

We should also write some tests and refactor dev_tools a bit, I thought about having some sort of module in charge of managing the urls. commands and dev_tools would depend on it. I found it a bit confusing to modify after a while not looking at the code tbh (even when I wrote it)

lua/flutter-tools/dev_tools.lua
akinsho: Just validated that `xdg-open` is *Not* valid on macOS so we need to change this to a variable which depends on the user's os and users either `xdg-open` on linux and `open` on mac and frankly I have no idea what on windows. I'd be fine with omitting windows for now if neither of us have access to a windows machine

akinsho (changes requested):
Just commented regarding `xdg-open` needing to be changed to an os based variable.

akinsho:
@Charly6596 adding tests sounds great, I've only recently added the workflow for it but would be good to start having tests with new functionality but since it's quite new it's not a hard blocker. Regarding refactoring it, tbh I'd rather avoid a whole new file just for url management which is related to dev-tools. I think all that can be handled in that module and if it becomes a general problem then factored out for other modules usage. Regarding a refactor of the module I don't really do much in that file so haven't really noticed it's current state 😆

lua/flutter-tools/dev_tools.lua
Charly6596: @akinsho I've copied that line from your config and it should be working now. _works on my machine_

lua/flutter-tools/dev_tools.lua
```diff
@@ -26,23 +26,30 @@ local devtools_profiler_url = nil
...
+local function try_get_profiler_url_chrome(data)
+    return data:match("(ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
 end
 
+local function start_browser()
+  local auto_open_browser = require("flutter-tools.config").get("dev_tools").auto_open_browser
+  if not auto_open_browser then return end
+  local url = M.get_profiler_url()
+  if url then
+    local open_cmd = vim.loop.os_uname() == "Darwin" and "open" or "xdg-open"
```
akinsho, line 47: This check is incorrect, `vim.loop.os_uname` returns a table that looks like
```lua
{
  machine = "x86_64",
  release = "20.5.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64"
}
```
so you need to check against the `sysname`. It should be
```suggestion
    local open_cmd = vim.loop.os_uname().sysname == "Darwin" and "open" or "xdg-open"
````

akinsho (changes requested):
The OS check for the open command isn't right unfortunately, I've made a comment/suggestion on how to fix it

akinsho (approved):
Thanks for the fix @Charly6596, seems to work fine now on my mac, will give it one last go on linux but otherwise looks good to merge.

akinsho:
Ok given this a try on Mac and Linux and seems to work fine. One thing to note is that when we close the application the debugger shows an error modal since it seems to expect the application to still be there, so I think we need to close the debugger before closing the app.

Issue #63, opened 2021-07-01
Add the ability to change the highlight groups.


ashishbinu:
I would like to change the colors for this notification window using highlight groups. Is there some highlight groups that i can achieve to do that.

![image](https://user-images.githubusercontent.com/56691622/124154423-1ae2bb80-dab3-11eb-963a-44cac0fcfabe.png)

akinsho:
@ashishbinu the plugins highlights are inherited from your nvim configurations/colorschemes highlights
In this case it is using `FloatBorder` and `NormalFloat` I would check what you have set those values to and tweaking them should allow you to control how these windows are coloured.

ashishbinu:
> @ashishbinu the plugins highlights are inherited from your nvim configurations/colorschemes highlights
> In this case it is using `FloatBorder` and `NormalFloat` I would check what you have set those values to and tweaking them should allow you to control how these windows are coloured.

I tried changing the background and forefground of the text using `NormalFloat` but it doesn't seem to change it. It changed the floating windows color in packer though.

akinsho:
@ashishbinu it depends on when you are trying to make this change are you setting the value of `NormalFloat` before this plugin is loaded. Can you check the value of `:verbose highlight NormalFloat` and paste it here? I've tweaked this highlight in my config, and it works as expected

ashishbinu:
@akinsho 
![image](https://user-images.githubusercontent.com/56691622/124169718-2f7b7f80-dac4-11eb-8a3c-c1a9f4a714a2.png)
This is what the command says about the NormalFloat

![image](https://user-images.githubusercontent.com/56691622/124169630-11158400-dac4-11eb-9fa4-7090fa7736e4.png)
This is where I have added the highlights to NormalFloat. 

It also highlights other stuff too...Even it worked , how can I only limit it to that notification only.

akinsho:
@ashishbinu I've linked you in a PR, can you please try that branch. FWIW, you are applying the highlighting too late in your example, you have applied the highlight *after* loading the plugin.

akinsho:
fixed by #64

PR #64, opened 2021-07-01
feat: create overridable notification highlights

akinsho:
This PR allows a user to override highlights to specifically change the colour of flutter tool notifications. Currently, this is only possible by changing global highlight groups

The groups are 
* `FlutterNotificationNormal` - content of the notification
* `FlutterNotificationBorder` - border of the notification

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,45 @@ EOF
 require("flutter-tools").setup{} -- use defaults
 ```
 
+# Functionality
+
+#### Run flutter app with hot reloading
+
+![hot reload](./.github/hot_reload.gif)
+
+#### Start emulators or connected devices
+
+![flutter-devices](https://user-images.githubusercontent.com/22454918/112320203-b5f31a80-8ca6-11eb-90b8-9ac934a842da.png)
+
+#### Visualise logs
+
+![dev log](./.github/dev_log.png)
+
+#### Widget guides (experimental, default: disabled)
+
+![Widget guides](./.github/outline_guide.png)
+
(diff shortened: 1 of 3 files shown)


akinsho:
@ashishbinu can you try using the highlight groups in this PR to see if they work for you.

akinsho:
fixes #63

ashishbinu:
> @ashishbinu can you try using the highlight groups in this PR to see if they work for you.

How can I pull this PR via packer?

ashishbinu:
image
It is working after adding the highlight commands in the setup function in packer.nvim.

image

ashishbinu:
@akinsho Wouldn't it be better to name the `FlutterNotificationNormal` to `FlutterNotficationContent`.

akinsho:
@ashishbinu no not really because the naming follow neovim highlighting convention, if you check the output of `:highlight<CR>` you will see that neovim uses normal for the content of the buffer which is what this is supposed to represent.

Can you clarify if this solves your issue?

akinsho:
Sorry, just noticed your message confirming it was working 👍🏾. Ok I'll update the README and merge this in soon.

ashishbinu:
> @ashishbinu no not really because the naming follows neovim's highlighting convention, if you check the output of `:highlight<CR>` you will see that neovim uses normal for the content of the buffer which is what this is supposed to represent.

If that's the case it seems right to have that name.

>Ok I'll update the README and merge this in soon.

Thanks for this PR.

PR #71, opened 2021-07-04
feat: add commandline args for run command

### LINKED ISSUE #70: FlutterRun custom flags
I'm working on an app with different `main`s to support various flavors (development and production).
I need the `--target` and `--flavor` flags.

I could implement this with a `flags` variable in `commands.lua` and making a command to set its value. Suggestions welcomed.

akinsho:
Add the ability to allow passing commands to flutter run. This is done by taking args via the command line command e.g.
`FlutterRun --flavor dev`
`FlutterRun --no-sound-null-safety`

This means this can also be mapped for reuse e.g.
`nnoremap ,frd <Cmd>FlutterRun --flavor dev`

fixes #70

### DIFF
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -3,7 +3,11 @@ local M = {}
 local function setup_commands()
   local utils = require("flutter-tools.utils")
   -- Commands
-  utils.command("FlutterRun", [[lua require('flutter-tools.commands').run()]])
+  utils.command(
+    "FlutterRun",
+    [[lua require('flutter-tools.commands').run_command(<q-args>)]],
+    { nargs = "*" }
+  )
   utils.command("FlutterReload", [[lua require('flutter-tools.commands').reload()]])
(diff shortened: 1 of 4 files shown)


Charly6596 (changes requested):

Charly6596:
Just to highlight this (since my comment was marked as outdated): 
I get this in the dev log when running without args in another project `Target file "" not found`.

akinsho:
Ah sorry @Charly6596 yeah I saw the comment, haven't had time to look into that yet, I'm wondering if some empty args are being passed to flutter somehow

akinsho:
@Charly6596 think this should be fixed now, `<q-args>` was returning `""` when no flags were passed which flutter interpreted as a path that it was being given to run. This shouldn't happen now

Charly6596:
@akinsho Oh I stopped working rigth now and I did just shutdown the pc. I will check later, but looks like it should work 😄

Charly6596:
Just checked and everything works as expected :+1: thank you so much for working on this

PR #74, opened 2021-07-07
fix(telescope): guard against nil opts in Run command from telescope

nghialm269:
This PR fixes bug: `attempt to index local 'opts' (a nil value)` when executing Run command in Telescope,


akinsho:
Awesome thanks for the fix @nghialm269

PR #95, opened 2021-09-19
feat: Execute flutter commands in lsp project root

clement-buchart:
Hello, 

Currently `flutter run` and `flutter pub get/upgrade` commands are run in neovim `cwd` directory by default.
In my project, the flutter project in a subdir in the monorepo, which means these command fail.

I couldn´t think of any reason not to use the dartls server root_dir config (which is correctly computed by the plugin) as the `cwd` for these commands.

I'm fairly new to neovim so that may be an oversight on my part, please let me know.

Thank you.

### DIFF
--- a/lua/flutter-tools/lsp/init.lua
+++ b/lua/flutter-tools/lsp/init.lua
@@ -174,4 +174,17 @@ function M.attach()
   end)
 end
 
+function M.get_lsp_root_dir()
+  for _, buf in pairs(vim.fn.getbufinfo({ bufloaded = true })) do
+    if api.nvim_buf_get_option(buf.bufnr, "filetype") == FILETYPE then
+      local clients = lsp.buf_get_clients(buf.bufnr)
+      for _, client in ipairs(clients) do
+        if client.config.name == SERVER_NAME then
+          return client.config.root_dir
+        end
+      end
+    end
+  end
+end
+
 return M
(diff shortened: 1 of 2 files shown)


akinsho:
@clement-buchart thanks for the PR 👍🏿 this looks good to me. Tbh the reason it wasn't like this from the start is that I've always had plugins that set vim's cwd to my project root so running the flutter command in cwd has always been fine for me but it's good to have it work more explicitly.

lua/flutter-tools/lsp/init.lua
```diff
@@ -174,4 +174,17 @@ function M.attach()
   end)
 end
 
+function M.get_lsp_root_dir()
```
akinsho, line 177: This looks like a duplication of the logic higher up in this page, It'd be great to instead generalise that function and re-use it in both places if possible

akinsho (commented):
@clement-buchart thanks for the PR 👍🏿 this looks good to me. Tbh the reason it wasn't like this from the start is that I've always had plugins that set vim's cwd to my project root so running the flutter command in cwd has always been fine for me but it's good to have it work more explicitly.

I've just added one comment about a function which looks like some code duplication I'd rather avoid and instead refactor into a common helper

lua/flutter-tools/lsp/init.lua
akinsho: I pushed a commit to take care of this 👍🏿

akinsho:
Thanks for the PR @clement-buchart 👍🏿

clement-buchart:
Oh, really sorry I didn't find time to make the adjustments. 

Thanks for taking care of it.

PR #96, opened 2021-10-06
feat(labels): Update to use extmarks

simrat39:
Also make sure the hl_mode is combine so it doesnt stand over the cursorline


akinsho:
@simrat39 thanks for catching this and sending a PR 👍🏿

PR #99, opened 2021-10-15
feat: support for fvm

rlch:
Hey @akinsho!

Thanks for this incredible package, I've been using it very happily for ~3 months. This is the plugin that gave me the final push away from VSC, so I have a lot to thank you for!  

One thing I've been missing out on from VSC is the ability to quickly switch Flutter SDK's. I frequently work with repos that run on different SDK's, + I have a local fork of Flutter that I'll occasionally use as well.

I've started doing some initial setup for integrating [FVM](https://fvm.app/), but I'd like to get your advice + support on a few things instead of arbitrarily tinkering with things I don't fully understand.

Right now I'm getting the following error `Client 1 quit with exit code 255 and signal 0` when attempting to attach the LSP to a buffer. I've verified that the paths to the analysis snapshot, dart_sdk and flutter binary are correct (point to the local FVM paths), and I've resolved the symlink so that it points exactly to the Flutter SDK in the FVM versions directory. So;
- Is there anything blatantly wrong I'm doing / anything left unaccounted for when setting up the references to the SDK paths?
- What's your preferred way of debugging + logging information in nvim development? I've noticed `print()` doesn't work in these functions.

**TODO's:**
- [x] Ensure FVM is only used if an SDK exists for the current project
- [x] Fall back to Flutter binary in path if `.fvm/` doesn't exist 
- [x] Fix error with attaching the LSP to a buffer
- [x] Clean up, document + add tests where necessary
- [x] Add separate Telescope menu for changing Flutter SDK + viewing selected (I think this will be really useful!) 
- [x] Disable FVM by default (?)

Sorry for all these questions, I'm an overt beginner to (n)vim development + Lua.
Any advice + support is kindly appreciated :smile:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -13,19 +13,23 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - run: date +%F > todays-date
-      - name: Restore cache for today's nightly.
+
+      - name: Get neovim v0.5.1
         uses: actions/cache@v2
         with:
-          path: _neovim
-          key: ${{ runner.os }}-x64-${{ hashFiles('todays-date') }}
+          path: build/neovim/v0.5.1
+          key: ${{ runner.os }}-appimage-0.5.1
 
-      - name: Prepare plenary
+      - name: Fetch dependencies
         run: |
           git clone --depth 1 https://github.com/nvim-lua/plenary.nvim ~/.local/share/nvim/site/pack/vendor/start/plenary.nvim
           ln -s $(pwd) ~/.local/share/nvim/site/pack/vendor/start
+
       - name: Run tests
         run: |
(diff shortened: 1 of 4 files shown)


akinsho:
@rlch this looks like it's shaped up nicely, will try and review properly tomorrow or early next week + the dependent PR. Regarding the `print` not working one potential issue could be that if your `set shortmess` includes option `F` this means that print statements called during an autocommand won't be printed since all output in an autocommand is silenced if that option is set

akinsho:
@rlch so I tried this out and it works great. One thing I noticed which I'm not sure if it's a blocking issue or not is that I actually didn't enable set the `fvm` key in my config and it worked anyway which I wouldn't have expected since it wasn't enabled, *but* I did manually use the picker to set the fvm version so maybe this is fine behaviour. In any case I'm happy to merge this just sitting on it because I don't want to break anyones setup 😆

PR #100, opened 2021-10-16
fix: prefer root-sdk binary over path

rlch:
Hey!

This PR is to fix weird issues I've come across running M1 under the `arm64` arch:
```
[ERROR][2021-10-16 14:25:45] .../vim/lsp/rpc.lua:401	"rpc"	"/opt/homebrew/Cellar/dart/2.14.4/libexec/bin/dart"	"stderr"	"Snapshot not compatible with the current VM configuration: the snapshot requires 'release no-code_comments no-dwarf_stack_traces_mode lazy_async_stacks lazy_dispatchers use_bare_instructions no-dedup_instructions no-\"asserts\" \"use_field_guards\" \"use_osr\" x64-sysv null-safety' but the VM has 'release no-code_comments no-dwarf_stack_traces_mode lazy_async_stacks lazy_dispatchers use_bare_instructions no-dedup_instructions no-\"asserts\" \"use_field_guards\" \"use_osr\" arm64-sysv null-safety'\n"
```

Here, the LSP is expecting an `x64` binary of Dart, but the VM was run with `arm64`. The reason for this mismatch is the following:
```lua
    config.cmd = config.cmd
      or {
        executable.dart_bin_name,
        analysis_server_snapshot_path(root_path),
        "--lsp",
      }
``` 
We should be using the Dart binary defined in `root_path`, instead of the binary exposed in PATH. That way, we can assert that the Dart SDK will be the same arch + version as defined in `pubspec.yaml`.

Thanks!

### DIFF
--- a/lua/flutter-tools/executable.lua
+++ b/lua/flutter-tools/executable.lua
@@ -45,7 +45,7 @@ end
 
 local function _flutter_sdk_dart_bin(flutter_sdk)
   -- retrieve the Dart binary from the Flutter SDK
-  return path.join(flutter_sdk, "dart", "bin")
+  return path.join(flutter_sdk, "bin", "dart")
 end
 
 ---Get paths for flutter and dart based on the binary locations


akinsho:
Thanks for the fix @rlch LGTM 👍🏿

PR #105, opened 2021-11-08
refactor: use vim.ui.input to prompt for name

nghialm269:
This PR changes `vim.fn.input` to `vim.ui.input` to allow custom ui for the prompt.

### DIFF
--- a/lua/flutter-tools/lsp/commands.lua
+++ b/lua/flutter-tools/lsp/commands.lua
@@ -4,25 +4,51 @@ function M.refactor_perform(command, ctx)
   local client = vim.lsp.get_client_by_id(ctx.client_id)
 
   local prompt
+  local default
   local kind = command.arguments[1]
   if kind == "EXTRACT_METHOD" then
     prompt = "Enter a name for the method: "
+    default = "newMethod"
   elseif kind == "EXTRACT_WIDGET" then
     prompt = "Enter a name for the widget: "
+    default = "NewWidget"
   elseif kind == "EXTRACT_LOCAL_VARIABLE" then
     prompt = "Enter a name for the variable: "
+    default = "newVariable"
   else
     client.request("workspace/executeCommand", command)
     return
   end
 
-  local optionsIndex = 6
-  local name = vim.fn.input(prompt)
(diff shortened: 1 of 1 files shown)


lua/flutter-tools/lsp/commands.lua
```diff
@@ -16,13 +16,24 @@ function M.refactor_perform(command, ctx)
...
+
+    local optionsIndex = 6
+    command.arguments[optionsIndex] = {
+      name = name,
+    }
+
+    client.request("workspace/executeCommand", command)
+  end
+
+  vim.ui.input(opts, on_confirm)
```
akinsho, line 36: @nghialm269 this will not exist for anyone not on nightly nvim so will cause breakage can you add a guard statement to check for the existence of `vim.ui.input` and if it exists use it, otherwise use the default input function

lua/flutter-tools/lsp/commands.lua
```diff
@@ -16,13 +16,24 @@ function M.refactor_perform(command, ctx)
...
+    prompt = prompt,
   }
 
-  client.request("workspace/executeCommand", command)
+  local on_confirm = function(name)
+    if name == nil then
+      return
+    end
+
+    local optionsIndex = 6
```
akinsho, line 28: I didn't notice this before but can you add a comment to explain why this value is 6 (rather than any other number) I'm sure it relates to the structure of some response or something but would be good to see something explaining that

akinsho (changes requested):
Looks good @nghialm269 , I've just left a comment about adding a check to ensure `vim.ui.input` exists before using it so this doesn't break for anyone on stable nvim or an older nightly

lua/flutter-tools/lsp/commands.lua
nghialm269: @akinsho I added a comment to explain, please tell me if it's not clear enough (or suggest me an explanation since I'm not good at this :D)

akinsho: That looks great @nghialm269

akinsho (approved):
Awesome LGTM 👍🏿 thanks for the fixes

PR #107, opened 2021-11-14
feat: enable code_actions execute without Outline window

leoncruz:
This create a simple command that shows a popup menu with code actions, like Outline window, but in source code, then users can run actions like: "Extract Widget" without open Outline window

### DIFF
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -19,6 +19,7 @@ local function setup_commands()
   --- Outline
   utils.command("FlutterOutlineOpen", [[lua require('flutter-tools.outline').open()]])
   utils.command("FlutterOutlineToggle", [[lua require('flutter-tools.outline').toggle()]])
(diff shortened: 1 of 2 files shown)


akinsho:
@leoncruz so this plugin's functionality is specific to just flutter rather than providing general lsp functionality like code actions. I only created that UI for the specific usecase of using with the outline window. I highly recommend using something like `telescope`, `lspsaga`, `nvim-code-action-menu` or `lsp-fastaction` which are dedicated plugins for code action which this plugin is not, so its implementation will not be as good as those.

I don't want users to expect this plugin to provide generalized lsp UI

akinsho:
I hope the rationale is clear, UI for renaming or code actions or lsp references are all general to all languages that support lsp. This plugin's remit/scope is just things that are for only flutter/dart so rather than a user trying to use this plugin *only* and expecting it to do *all* the lsp things which is a huge amount of work this plugin only does things that make working with flutter or dart specifically easier/possible

leoncruz:
Hi, @akinsho, thanks for your feedback!

Issue #91, opened 2021-08-22
[Bug]: Can't hot reload or hot restart with DAP


DanielCardonaRojas:
Can't hot reload or hot restart when running the flutter through nvim-dap,
the documentation says that  when debugging is enabled  the app should be ran with `:lua require('dap').continue()<CR>`. 

This results in a log message: `[Flutter-tools]: Sorry! fFutter is not running`

akinsho:
Hi @DanielCardonaRojas thanks for reporting this. There are a few issues with using DAP I still need to work out, which require a bit of know how regarding dap's internals I don't have quite yet. Will leave this open till I or anyone else who might want to contribute has time to look into a good solution.

FWIW I think the issue is that dap is starting the app itself when continue is used which means that this plugin has no real connection to the process it creates and no way of knowing that it even exists or how to interact with it.

@DanielCardonaRojas you might be able to get this work by starting your app and using the `attach` option that should be presented to you when you hit continue instead of other option (I forget the name).

I think the solution lies with only providing an attach option so that this plugin can create and manage the process and dap just attaches to the process we are managing

clement-buchart:
Hello @akinsho, 

I've been poking around at this and I thought you (or people following this issue) might be interested.

- The process you describe (starting the app via the `flutter-tools.nvim` and attaching to the process) is achievable and does work with 2 caveats :
    - It seems to work only for hot reload, for hotRestart I get
```
Performing hot restart...                                       
E/flutter ( 8452): [ERROR:flutter/runtime/dart_isolate.cc(170)] Could not prepare isolate.
E/flutter ( 8452): [ERROR:flutter/runtime/runtime_controller.cc(382)] Could not create root isolate.
Restarted application in 1,003ms.
Application finished.
hot restart failed to complete: _flutter.runInView: (-32000) Could not run configuration in engine.
#0      new _OutstandingRequest (package:vm_service/src/vm_service.dart:1626:45)
#1      VmService._call (package:vm_service/src/vm_service.dart:2109:21)
#2      VmService.callServiceExtension (package:vm_service/src/vm_service.dart:2080:14)
#3      VmService.callMethod (package:vm_service/src/vm_service.dart:2062:12)
#4      FlutterVmService.callMethodWrapper (package:flutter_tools/src/vmservice.dart:451:28)
#5      FlutterVmService.runInView (package:flutter_tools/src/vmservice.dart:530:11)
<asynchronous suspension>
[...]

PR #108, opened 2021-11-21
feat: Introduce dap-based runner

clement-buchart:
Fixes https://github.com/akinsho/flutter-tools.nvim/issues/91

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -4,23 +4,45 @@ local utils = require("flutter-tools.utils")
 local devices = require("flutter-tools.devices")
 local config = require("flutter-tools.config")
 local executable = require("flutter-tools.executable")
-local dev_log = require("flutter-tools.log")
 local dev_tools = require("flutter-tools.dev_tools")
 local lsp = require("flutter-tools.lsp")
+local job_runner = require("flutter-tools.runners.job_runner")
+local dap_runner = require("flutter-tools.runners.dap_runner")
+local dap_ok, dap = pcall(require, "dap")
 
 local M = {}
 
----@type Job
-local run_job = nil
 ---@type table
 local current_device = nil
+
+---@class FlutterRunner
+---@field is_running fun():boolean
+---@field run fun(paths:table, args:table, cwd:string, on_run_data:fun(is_err:boolean, data:string), on_run_exit:fun(data:string[]))
+---@field cleanup fun()
+---@field send fun(cmd:string)
+
+---@type FlutterRunner
+local runner = nil
+
+function M.use_dap_runner()
+  local dap_requested = config.get("debugger").run_via_dap
+  if dap_requested then
+    if not dap_ok then
+      utils.notify("dap runner was request but nvim-dap is not installed!\n" .. dap, utils.L.ERROR)
+      return false
+    end
+    return true
+  else
+    return false
+  end
+end
 
 function M.current_device()
   return current_device
 end
 
 function M.is_running()
-  return run_job ~= nil
+  return runner ~= nil and runner:is_running()
 end
 
 local function match_error_string(line)
@@ -50,18 +72,19 @@ end
 
 ---Handle output from flutter run command
 ---@param is_err boolean if this is stdout or stderr
----@param opts table config options for the dev log window
----@return fun(err: string, data: string, job: Job): nil
-local function on_run_data(is_err, opts)
-  return vim.schedule_wrap(function(_, data, _)
-    if is_err then
-      ui.notify({ data })
-    end
-    if not match_error_string(data) then
-      dev_tools.handle_log(data)
-      dev_log.log(data, opts)
-    end
-  end)
+local function on_run_data(is_err, data)
+  if is_err then
+    ui.notify({ data })
+  end
+end
+
+local function shutdown()
+  if runner ~= nil then
+    runner:cleanup()
+  end
+  runner = nil
+  current_device = nil
+  dev_tools.on_flutter_shutdown()
 end
 
 ---Handle a finished flutter run command
@@ -80,12 +103,7 @@ local function on_run_exit(result)
       end,
     })
   end
-end
-
-local function shutdown()
-  run_job = nil
-  current_device = nil
-  dev_tools.on_flutter_shutdown()
+  shutdown()
 end
 
 --- Take arguments from the commandline and pass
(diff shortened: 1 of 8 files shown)


akinsho:
@clement-buchart this looks great haven't had time yet to go through the PR but will run it locally for a bit and when I get a chance feedback on the PR as well but did a quick scheme and I like the approach generally 🙌🏿

lua/flutter-tools/commands.lua
```diff
@@ -4,23 +4,45 @@ local utils = require("flutter-tools.utils")
 local devices = require("flutter-tools.devices")
 local config = require("flutter-tools.config")
 local executable = require("flutter-tools.executable")
-local dev_log = require("flutter-tools.log")
 local dev_tools = require("flutter-tools.dev_tools")
 local lsp = require("flutter-tools.lsp")
+local job_runner = require("flutter-tools.runners.job_runner")
+local dap_runner = require("flutter-tools.runners.dap_runner")
+local dapSuccess, dap = pcall(require, "dap")
```
akinsho, line 11: Variables should be snake case instead of camel e.g. `dap_success` also I prefer the term `ok` e.g. `dap_ok` very minor point though 🤷🏿‍♂️

lua/flutter-tools/commands.lua
```diff
@@ -4,23 +4,45 @@ local utils = require("flutter-tools.utils")
...
+---@field is_running fun():boolean
+---@field run fun(paths:table, args:table, cwd:string, on_run_data:fun(is_err:boolean, data:string), on_run_exit:fun(data:string[]))
+---@field cleanup fun()
+---@field send fun(cmd:string)
+
+---@type FlutterRunner
+local runner = nil
+
+function M.use_dap_runner()
+  local dapRequested = config.get("debugger").run_via_dap
```
akinsho, line 28: Similar to the above regarding casing, ideally all variables should be snake case

lua/flutter-tools/commands.lua
```diff
@@ -120,34 +140,23 @@ function M.run(opts)
...
-      on_stderr = on_run_data(true, conf),
-      on_exit = vim.schedule_wrap(function(j, _)
-        on_run_exit(j:result())
-        shutdown()
-      end),
-    })
-
-    run_job:start()
+    current_device = device
+    if M.use_dap_runner() then
```
akinsho, line 145: this could be represented with a ternary i.e. `runner = M.use_dap_runner() and dap_runner or job_runner`

lua/flutter-tools/runners/dap_runner.lua
```diff
@@ -0,0 +1,138 @@
...
+      for line in body.output:gmatch("[^\r\n]+") do
+        if not started then
+          table.insert(before_start_logs, line)
+        end
+        on_run_data(body.category == "sterr", line)
+      end
+    end
+  end
+
+  local handleTermination = function()
```
akinsho, line 43: similar casing issue as above

akinsho (commented):
@clement-buchart thanks for your work on this, it looks great 👍🏿 I'm running it locally just to see how it behaves but in the meantime I've made a few comments, primarily around code style.

clement-buchart:
Sorry for the rookie mistakes, my lua brain isn't completely formed yet ^^
I think I fixed all the casing issues.

lua/flutter-tools/runners/dap_runner.lua
```diff
@@ -0,0 +1,138 @@
...
+      launch_configurations,
+      "Select launch configuration",
+      function(item)
+        return fmt("%s : %s", item.name, item.program, vim.inspect(item.args))
+      end
+    )
+  end
+  launch_config = vim.deepcopy(launch_config)
+  launch_config.cwd = cwd
+  launch_config.args = vim.list_extend(launch_config.args, args)
```
akinsho, line 102: I hit an error on this line saying that `dst` is empty, presumably `args` can be nil and is not handled
```
E5108: Error executing lua vim/shared.lua:361: dst: expected table, got nil
stack traceback:
        [C]: in function 'error'
        vim/shared.lua:608: in function 'validate'
        vim/shared.lua:361: in function 'list_extend'
        ...tter-tools.nvim/lua/flutter-tools/runners/dap_runner.lua:102: in function 'run'
        .../local-flutter-tools.nvim/lua/flutter-tools/commands.lua:150: in function 'get'
        .../local-flutter-tools.nvim/lua/flutter-tools/commands.lua:126: in function 'run'
        .../local-flutter-tools.nvim/lua/flutter-tools/commands.lua:114: in function 'run_command'
        [string ":lua"]:1: in main chunk
```

clement-buchart: Ah seems like dst is the first arg : 
`list_extend({dst}, {src}, {start}, {finish})`
So it must be `launch_config.args` that is nil (indeed `args` is initialized as `{}` in commands.lua).

My project always set some --target so I failed to catch this

I'll push a fix thanks.

akinsho:
@clement-buchart I've noticed whilst running this that the flutter log buffer isn't created or populated whilst running with `dap` is this by design? I think that buffer is quite a nice tool as it can be syntax highlighted and is a good base that I envision building functionality for eventually.

clement-buchart:
Yes, I moved the function sending lines to be printed in this buffer in the JobRunner because it seemed redundant with the dap-repl buffer. 

Initially, I had both working so it shouldn't be too hard puting it back in the common part, if you prefer.

akinsho:
@clement-buchart I like having a log buffer that this plugin controls, since the dap repl can't really be syntax highlighted properly since it doesn't contain just log output but prompts as well. Also, I've been toying with the idea of allowing jumping to and from locations in that buffer by creating custom mappings for that buffer. In short, I think it's fine if a user wants to use dap's UI for the output, but there should always be a flutter log, and they can just disable it if they want.

clement-buchart:
Ok I'll put it back no problem

akinsho:
Okay tried this for a couple of days and it seems to work fine 👍🏿 Thanks a bunch for your work on this @clement-buchart

akinsho:
@clement-buchart sorry pulling you back into this. I've recently started working with `run_via_dap` enabled and noticed something odd and am wondering if you encountered it when you implemented this or are currently seeing it, when I start an app via dap the lsp seems to go out of whack i.e. there are a bunch of spurious diagnostics, widget guides are on the wrong lines etc. It definitely seems to be due to some interaction, since it only happens when running via dap. I'm wondering if you've seen this before?

I'm trying to figure out if this is my setup/environment or some strange interaction between how the debugger is run and the dart language server runs

clement-buchart:
Hello,

Nope sorry, never encountered symptoms such as this.

akinsho:
@clement-buchart any chance you can share the configuration you use with dap, the key might be in there somewhere 🙏🏿

PR #110, opened 2021-12-02
fix(daprunner): Check for nil in dap events

clement-buchart:


akinsho:
Thanks for the quick fix @clement-buchart 🥇

Issue #111, opened 2021-12-05
Support: How do you detach from a running flutter process?


joshpetit:
Hey, thanks for making this awesome plugin :smile: 

I was wondering if it was possible to detach from the running flutter process? This is something that's capable in coc-flutter

akinsho:
@joshpetit you can quit/close use the `FlutterQuit` command to close the running application. All the available commands should be documented in the README, although I might have forgotten one or two

joshpetit:
@akinsho Yea, `FlutterQuit` works fine. `FlutterDetach` would stop the process from running inside the terminal but leave it running on the phone. It may just be something coc-flutter implemented so doesn't really need to be in here. The command is called `flutter.dev.detach` and described as "Detach server".

akinsho:
Ah thanks for clarifying @joshpetit that is different functionality. I never used that whilst using `coc-flutter` what would the use case for that be? I definitely haven't implemented all the commands it does, primarily because I only wanted to add stuff that had a clear benefit. If it's native to the flutter binary then it shouldn't be too hard to add, although I guess before that it'd be good to determine if it will be useful to a lot of people.

joshpetit:
@akinsho , no problem. It left the app running on the phone. I often need to demo the app, so instead of having to start the build process again to get the debug apk installed, I can just detach the debug apk and demo that after I'm done working. If I detach it also makes the version that remains on the phone the most recent, since when I do flutter quit and reopen the app later from the launcher, it's usually a few iterations previous older. 

That's the main benefit, but if it's not a usecase you can see a lot of people using then there's not much reason to implement it honestly haha. I can see it being useful and also something no one ever really uses :sweat_smile:

akinsho:
@joshpetit if you are up for submitting a PR to add that then I'm fine with the addition, personally I don't think I'd use it but could see how someone else might perhaps. In either case I'll accept it as a PR but need to focus on adding as general as possible functionality personally since I have limited time so probably won't add this myself.

joshpetit:
finished!

PR #113, opened 2021-12-27
feature: add flutter detach command

joshpetit:
#111 

done :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ require("flutter-tools").setup{} -- use defaults
 - `FlutterReload` - Reload the running project.
 - `FlutterRestart` - Restart the current project.
 - `FlutterQuit` - Ends a running session.
+- `FlutterDetach` - Ends a running session locally but keeps the process running on the device.
 - `FlutterOutlineToggle` - Toggle the outline window showing the widget tree for the given file.
 - `FlutterOutlineOpen` - Opens an outline window showing the widget tree for the given file.
 - `FlutterDevTools` - Starts a Dart Dev Tools server.
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -9,6 +9,7 @@ local function setup_commands()
     { nargs = "*" }
   )
   utils.command("FlutterLspRestart", 'lua require("flutter-tools.lsp").restart()')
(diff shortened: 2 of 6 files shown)


joshpetit:
Only thing I'm unsure about is the `command_requests` for the dap runner. not really sure what `detach` would be for it so I just put `detach = "detach"`. Would there be a place for me to verify this?

[The line im talking about](https://github.com/akinsho/flutter-tools.nvim/pull/113/commits/433f242e252ea451ecc2354143ad1498b49844cb#diff-9366d637f63c2f176a54a3e0af43f5e21f3aa6c97e72be09809d49d91711aee0R19)

akinsho:
@joshpetit thanks for working on this, generally looks fine to me. Regarding the `command_requests` @clement-buchart actually added the functionality and tbh I don't know how he figured out which names matched which for the dap runner. Will see if he can enlighten us, otherwise I'll merge this as a best effort we can try and improve on if it doesn't work

clement-buchart:
Hello, 

Looking at https://github.com/Dart-Code/Dart-Code/blob/f58ca38e451d81f4d0e20bbb6b8ebb06a66264a4/src/debug/run_daemon_base.ts#L104 there is a detach method, but it's only ever called if the dap configuration was started in attach mode, and we always start a "launch" configuration. 

So for it to work we would need either : 
- Add some codepath in Dart-Code to be able to call this method explicitly via a request and contribute this upstream.
- Refactor the dap-runner to use an attach configuration. This would imply actually starting the flutter process ourselves and then attaching to it. 
 
First solution is probably simpler to implement and "better", but dependent on another repo.

I personally will not have a lot of time to contribute it is the foreseeable future. 

May I suggest simply not adding anything to the command_requests of the dap runner ? 
This way it should fall back to this codepath which will indicate it's not implemented yet for this runner : 
```  
elseif not quiet then
    utils.notify("Command " .. cmd .. " is not yet implemented for DAP runner")
end
```
(thus current implementation n this PR will do nothing silently)

Cheers.

akinsho:
I agree with omitting the functionality for the dap runner as both alternative options are quite involved. Thanks for the input @clement-buchart 👍🏾

joshpetit:
@akinsho Just removed it 👍

akinsho:
Thanks for the PR @joshpetit just merged it 👍🏿

Issue #112, opened 2021-12-15
[Bug]: Error not shown during flutter run


joshpetit:
**Describe the bug**
<-- Please ensure that you have read the README.md -->
<-- Also report bugs using a minimal init.vim or init.lua rather than your full configuration -->
<-- Do Not Delete this template or will close the issue as it helps me debug reports -->

**To Reproduce**

1. Have an error in your dart file
2. Do FlutterRun
3. Don't See error in the output

**Expected behavior**
I expected the dev log to show the error and then quit

**Screenshots**
This is usually shown:

```
Launching lib/main.dart on Pixel 4a in debug mode...
Running Gradle task 'assembleDebug'...                          
Running Gradle task 'assembleDebug'...                             18.6s
```

It stops after assembleDebug and doesn't show the eventual error

**System information**

- OS:
- Flutter version: 2.8.0
- Is flutter in $PATH: yes
- neovim version: .6

**Additional context**
You can see the error when you run `flutter run` in the terminal

akinsho:
I believe this should now be fixed by #114

joshpetit:
Hm, this still does not seem to work for me. I updated and when running with an error it still just shows `Running Gradle task 'assembleDebug'`

PR #114, opened 2021-12-31
fix: callback name for plenary job execution

leoncruz:
This pull request fix the name of callback for plenary job. The incorrect name is causing errors like this issue: [#112](https://github.com/akinsho/flutter-tools.nvim/issues/112)


akinsho (approved):
LGTM 💯 thanks for catching this @leoncruz

PR #136, opened 2022-03-02
feat(commands): add widget inspector and construction lines

bjornevik:
This adds a widget inspector and a construction lines toggle to the "Command Palette".

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -223,6 +223,16 @@ function M.copy_profiler_url()
   end
 end
 
+---@param quiet boolean
+function M.widget_inspector(quiet)
+  send("inspect", quiet)
+end
+
+---@param quiet boolean
+function M.construction_lines(quiet)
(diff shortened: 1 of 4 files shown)


lua/flutter-tools/commands.lua
```diff
@@ -223,6 +223,16 @@ function M.copy_profiler_url()
   end
 end
 
+---@param quiet boolean
+function M.widget_inspector(quiet)
+  send("inspect", quiet)
```
akinsho, line 228: These commands will also need to be added to the `DAPRunner` here
https://github.com/akinsho/flutter-tools.nvim/blob/c0909a073348db3dbc5bdcbeddd49cc6f04ebeb1/lua/flutter-tools/runners/dap_runner.lua#L16

akinsho (changes requested):
Thanks for the PR this LGTM except it also needs to be added to the command requests map in the `DAPRunner`

PR #143, opened 2022-04-24
fix(guides): do not draw guides over text

sidlatau:
Before:
image
After:
image


akinsho (approved):
@sidlatau thanks for the PR 👍🏾, I won't have time to test it myself but looks good to me.

PR #147, opened 2022-05-22
Hide Flutter dev log from buffers list.

divan:
Flutter dev log buffer is appearing in the buffer list with a filename '__FLUTTER_DEV_LOG__'. It's handy to keep that buffer visible (in a split window), but it causes problems when cycling through buffers or resizing it.

This PR sets 'buflisted' property to false, so the buffer is not listed.

This will probably require implementing some kind of log window toggling (discussed in https://github.com/akinsho/flutter-tools.nvim/issues/84).

In general, this echoes abovementioned issue in a way to bring more consistency between log and outline windows for this plugin.

### DIFF
--- a/lua/flutter-tools/log.lua
+++ b/lua/flutter-tools/log.lua
@@ -110,6 +110,7 @@ function M.__resurrect()
   vim.bo[buf].modifiable = false
   vim.bo[buf].modified = false
   vim.bo[buf].buftype = "nofile"
+  vim.bo[buf].buflisted = false
 end
 
 function M.clear()
--- a/lua/flutter-tools/utils/init.lua
+++ b/lua/flutter-tools/utils/init.lua
@@ -46,7 +46,7 @@ function M.buf_valid(bufnr, name)
   if bufnr then
     return api.nvim_buf_is_loaded(bufnr)
   end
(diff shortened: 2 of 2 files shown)


akinsho:
@divan thanks for the PR, so this is actually an intentional choice to keep it visible in the list, primarily because people regularly do thing like copying from the log or just needing it to be visible to refer back to. I think as you say if there was a consistent way to toggle it reliably in all cases then it wouldn't be needed, but it feels like this should come after that. 
If this is taken away now without a toggle function, users will not be able to find the buffer without a bunch of faff.

divan:
@akinsho totally agree. I'll try to work a bit on toggling. So far my understanding is that split and tabs modes are quite different.  Split mode should be identical to outline (UX-wise). Tab mode is completely different.
If that's correct, than the challenge is to distinguish those to modes and apply different set of buffer configurations.

akinsho:
@divan I would actually approach more simply and regardless of a user's configuration if they are toggling then it will be a split, but the command should be named as such e.g. toggle split. I would basically open a split window and track the ID and so if they hit toggle again and the window is still open in any tabpage then close it. If it isn't open in any tabpage then open it in the current one. I guess you'll also need to keep track of the buffer number which should be consistent regardless of the tab, so you can open it when toggling it open

sidlatau:
Already added.

PR #154, opened 2022-06-13
chore(readme): mention of excluding pub packages on windows

RobertBrunhage:
# Hi 👋

Regarding the topic discussed in https://github.com/akinsho/flutter-tools.nvim/issues/57 this fix will add another default path to prevent the LSP to load when navigating package code.

Feel free to let me know if there is another way to do the change!

## Visual example of the path

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,11 @@ you jump to the definition of `StatelessWidget`, the lsp will not try and index 
 files in that directory. If for some reason you would like this behaviour set `analysisExcludedFolders = {}`
 You cannot/should not edit the files in the sdk directly so diagnostic analysis of these file is pointless.
 
+**Exclude Note for Windows:**
+To ignore packages installed with pub, consider adding `vim.fn.expand("$HOME/AppData/Local/Pub/Cache")` to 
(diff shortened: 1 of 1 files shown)


akinsho:
@RobertBrunhage thanks for the PR, tbh I have worked very hard to avoid OS specific logic in this code base given it's used cross-platform, do you know if there is a way of deriving this path for all systems? Like the output of a flutter command or an environment variable. As I think it'd problematic to add this just for Windows users and would advise instead to put this into people's configs

akinsho:
Not familiar with windows at all but this looks kind of similar to the `XDG_CACHE` env var

RobertBrunhage:
@akinsho understandable, sadly I am not aware if they have a variable for the path that is applicable for the different platforms. I could update the PR to just mention it in the README instead if wanted, or we can close 😊

akinsho:
I think a mention in the README would be great 👍🏿

RobertBrunhage:
Updated the README, let me know if you want any changes 👏

akinsho:
LGTM 👍🏿

PR #156, opened 2022-06-20
feat(commands): open dev tools

huylg:
Hi, I think opening dev tools directly is more convenient than copying profile url and opening browser manually.

### DIFF
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -30,6 +30,10 @@ local function setup_commands()
     "FlutterOpenDevTools",
     [[lua require('flutter-tools.commands').open_dev_tools()]]
   )
+  utils.command(
+    "FlutterGenerate",
+    [[lua require('flutter-tools.commands').generate()]]
+  )
 
   --pub
   utils.command("FlutterPubGet", [[lua require('flutter-tools.commands').pub_get()]])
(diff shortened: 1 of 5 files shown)


huylg:
hi @akinsho, please take a look at my pr :D

lua/flutter-tools/commands.lua
```diff
@@ -223,6 +223,11 @@ function M.copy_profiler_url()
   end
 end
 
+function M.open_dev_tools(quiet)
+  send("open_dev_tools",quiet)
```
akinsho, line 227: @huylg this isn't handled in the debug runner, so anyone running this project via dap/the debugger won't be able to open this. Can you have a look in that file to see if there is command to trigger the devtools via dap

lua/flutter-tools.lua
```diff
@@ -26,6 +26,12 @@ local function setup_commands()
     "FlutterCopyProfilerUrl",
     [[lua require('flutter-tools.commands').copy_profiler_url()]]
   )
+  utils.command(
+    "FlutterOpenDevTools",
+    [[lua require('flutter-tools.commands').open_dev_tools()]]
+  )
+
+  --pub
```
akinsho, line 34:
```suggestion
  --- Pub
```

akinsho (commented):
Generally LGTM I've left a comment about trying to maintain feature parity between the job and debugger runner, so they both can open the dev tools. It's not a hard blocker if you can't find a command for it but I think testing some stuff out a bit would be good

lua/flutter-tools/commands.lua
huylg: I got it @akinsho, thanks for reviews.

PR #158, opened 2022-07-10
refactor(lsp): use vim.lsp.start if on nightly

akinsho:
Nvim 0.8 introduced `vim.lsp.start` which removes a lot of the hard work of manually starting a server. This PR lays the groundwork for using it, but only for those on nvim nightly for now

### DIFF
--- a/lua/flutter-tools/lsp/init.lua
+++ b/lua/flutter-tools/lsp/init.lua
@@ -127,7 +127,9 @@ function M.restart()
     client.stop()
     local client_id = lsp.start_client(client.config)
     for _, buf in pairs(bufs) do
-      lsp.buf_attach_client(buf, client_id)
+      if client_id then
+        lsp.buf_attach_client(buf, client_id)
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lua/flutter-tools/lsp/init.lua
```diff
@@ -194,32 +193,48 @@ local function get_server_config(user_config, callback)
...
+  local debug_log = create_debug_log(user_config.debug)
+  debug_log("attaching LSP")
+
+  -- FIXME: When nvim 0.8 is released remove the legacy_server_init
+  if vim.version().minor < 8 then
+    legacy_server_init(api.nvim_get_current_buf(), user_config)
+  else
+    local fs = vim.fs
+    get_server_config(user_config, function (c)
+      c.root_dir = fs.dirname(fs.find(ROOT_PATTERNS, { upward = true })[1])
```
changyiyao, line 234: c.root_dir should respect existing dartls's root_dir first as before.
`c.root_dir = M.get_lsp_root_dir() or fs.dirname(fs.find(ROOT_PATTERNS, { upward = true })[1])`
may be better.

akinsho: Feel free to raise a PR @changyiyao in the first instance `get_lsp_root_dir()` will be `nil` because there is no running server which is why I did it this way but as you say when there is it can respect the existing server's root, although frankly I don't understand why it would matter since the root is derived from the `fs.find` method in the first place so running it again should be fine 🤷🏿‍♂️

PR #161, opened 2022-07-11
fix: use correct command option name

sidlatau:


sidlatau:
Oh, I see I am too late :)

akinsho:
Thanks anyway though 👍🏾🙏🏾

PR #162, opened 2022-07-12
feat(command): Flutter attach

huylg:
hi @akinsho, Flutter attach is a valuable feature of flutter so I create this pr :)


huylg:
@akinsho please take a look when you have some time :D

lua/flutter-tools/commands.lua
```diff
@@ -105,12 +105,13 @@ end
...
   args = args and args ~= "" and vim.split(args, " ") or nil
-  M.run({ args = args })
+  M.run('run', { args = args })
 end
 
 ---Run the flutter application
+---@param action string
 ---@param opts table
-function M.run(opts)
+function M.run(action, opts)
```
rlch, line 114: Not sure we should be coupling the `attach` command with logic meant for the `run` command. IMO we should create a separate function for handling `attach`; and based off my little experience using it, maybe the logic could look something like this..?

```mermaid
stateDiagram-v2
    state explicit_device <<choice>>
    state running <<choice>>
    cd: Open choose device UI
    nanr: Notify app not running
    nout: Notify output
    rc: Run attach command w/ args
    [*] --> running
    running --> nanr
    nanr --> [*]
    running --> explicit_device : App running
    explicit_device --> cd : No device passed to command
    explicit_device --> rc
    cd --> rc
    rc --> nout
    nout --> [*]
```
- There should be existing functions to help you implement this

@akinsho You may want to verify the above; maybe this is overkill!

akinsho: @rlch I like the sounds of this approach because it is generally in keeping with the way I've tried to make things work in this plugin which is to make the flutter command used via this plugin "do the right thing" if there are multiple devices then offer a choice and use the one they choose. If something isn't started, ask to start it.

I haven't really looked too closely yet but definitely agree that shoe horning something in is not the best way to do this. It's good to have commands available but I don't want that to just be workarounds to get things working that don't try to offer an ideal user flow.

TLDR: I think a flow like the above would be a good way to implement this. I never use the attach command, so have no specific input other than I think it should be graceful and separate to run.

huylg: @rlch @akinsho it's clear to me, I will reimplement it. Thanks for review

PR #163, opened 2022-07-14
fix: attach LSP to buffer only once

sidlatau:
With the new LSP attach functionality on 0.8 I stated getting `dartls` attached to the buffer several times. Looks like the problem is that `ftplugin` is called several times.

### DIFF
--- a/ftplugin/dart/lsp.lua
+++ b/ftplugin/dart/lsp.lua
@@ -1,1 +1,5 @@
+if vim.b.did_ftplugin_dart_lsp then
+  return
+end
+vim.b.did_ftplugin_dart_lsp = true
 require('flutter-tools.lsp').attach()


akinsho:
Thanks for picking this up, so it's actually expected that ftplugin will be called for every new dart file or every time a `FileType` autocommand is triggered. I guess it by default always tries to re-attach when it is called if there is already a running server

akinsho (approved):
LGTM 👍🏿

changyiyao:
Hi guys, @sidlatau @akinsho 
I encounter that some dart buffers are not attach to the server( I don't know the root cause, which may occur when starting up with auto-session or viewing dart files in other outside dependent packages.) so that I must attach the buffer manually `:lua vim.lsp.buf_attach_client(0, 1)`. 
So I find out and revert this changes and everything seems go well now. Also I don't know what the problem duplicated `vim.lsp.start` calls lead to. If there is I think checking whether the buffer is attached to an existing active client via `vim.lsp.get_active_clients({name, bufnr})` may be the preferred solution.

PR #167, opened 2022-07-22
feat(setup): add pubspec.yaml as another start point

changyiyao:
Without this, I need to open any dart file before :FlutterPubGet and :FlutterRun when first entering neovim with pubspec.yaml.


akinsho (approved):
LGTM @changyiyao thanks

PR #169, opened 2022-07-22
feat(lsp): support dart/textDocument/super method

changyiyao:
Add a new command `FlutterSuper` to invoke the dart lsp method to navigate to its `super`.

### DIFF
--- a/ftplugin/dart/lsp.lua
+++ b/ftplugin/dart/lsp.lua
@@ -1,5 +1,1 @@
-if vim.b.did_ftplugin_dart_lsp then
-  return
-end
-vim.b.did_ftplugin_dart_lsp = true
 require('flutter-tools.lsp').attach()
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -60,6 +60,16 @@ local function setup_commands()
   cmd("FlutterLogClear", function()
     require("flutter-tools.log").clear()
   end, {})
+  --- LSP
+  local lsp = vim.lsp
+  cmd("FlutterSuper", function()
+    lsp.buf_request(
+      0,
+      "dart/textDocument/super",
+      lsp.util.make_position_params(),
+      lsp.handlers["textDocument/definition"]
+    )
+  end, {})
 end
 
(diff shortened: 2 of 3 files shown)


lua/flutter-tools.lua
```diff
@@ -60,6 +60,16 @@ local function setup_commands()
   cmd("FlutterLogClear", function()
     require("flutter-tools.log").clear()
   end, {})
+  --- LSP
+  local lsp = vim.lsp
+  cmd("FlutterSuper", function()
+    lsp.buf_request(
```
akinsho, line 66: @changyiyao can you move this logic to the lsp module I don't think logic like this should be done inline inside the flutter tools entry point file but rather inside the relevant module then imported here

ftplugin/dart/lsp.lua
```diff
@@ -1,5 +1 @@
-if vim.b.did_ftplugin_dart_lsp then
```
akinsho, line 1: Normally this should be in a separate PR since your changes don't mention this at all but it's fine since it's a small change

akinsho (changes requested):
LGTM generally but the function needs to be added to the right module and imported here instead.

lua/flutter-tools.lua
changyiyao: @akinsho Thanks, Let me have a try

akinsho:
@changyiyao I've rebased your changes and moved the logic into the lsp module on my side but can't push to your branch to update it as I was going to merge it once my changes were pushed. Anyway you don't need the additional changes your branch as I've added fixes in main for those. All you need to do now is just move the logic to `flutter-tools/lsp` and require in the function

```lua
-- In lsp.lua

function M.super()
  lsp.buf_request(
    0,
    "dart/textDocument/super",
    lsp.util.make_position_params(),
    lsp.handlers["textDocument/definition"]
  )
end
```
Then in `flutter-tools.lua`
```lua
  cmd("FlutterSuper", function()
    require("flutter-tools.lsp").super()
  end, {})
```

PR #170, opened 2022-07-27
feat(commands): add FlutterSuper and FlutterReanalyze

changyiyao:
FlutterSuper respects Spec: https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/tool/lsp_spec/README.md#darttextdocumentsuper-method

FlutterReanalyze respects Spec: https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md#dartreanalyze-method.

### DIFF
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -60,6 +60,10 @@ local function setup_commands()
   cmd("FlutterLogClear", function()
     require("flutter-tools.log").clear()
   end, {})
+  --- LSP
+  cmd("FlutterSuper", function()
+    require("flutter-tools.lsp").dart_lsp_super()
+  end, {})
 end
 
 ---Initialise various plugin modules
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -64,6 +64,9 @@ local function setup_commands()
   cmd("FlutterSuper", function()
     require("flutter-tools.lsp").dart_lsp_super()
   end, {})
+  cmd("FlutterReanalyze", function()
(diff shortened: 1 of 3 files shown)


lua/flutter-tools.lua
```diff
@@ -64,6 +64,9 @@ local function setup_commands()
   cmd("FlutterSuper", function()
     require("flutter-tools.lsp").dart_lsp_super()
   end, {})
+  cmd("FlutterReanalyze", function()
```
akinsho, line 67: I believe the documentation recommends not using this method. I actually historically omitted it on purpose for this reason. See the excerpt from the docs below

> Clients should not usually need to call this method - needing to do so may indicate a bug in the server.

akinsho (approved):
LGTM

lua/flutter-tools.lua
akinsho: 🤷🏿‍♂️ will leave it up to the user to use as they please and remove it if it introduces problems

akinsho:
Thanks for the contributions @changyiyao

lua/flutter-tools.lua
changyiyao: > I believe the documentation recommends not using this method. I actually historically omitted it on purpose for this reason. See the excerpt from the docs below
> 
> > Clients should not usually need to call this method - needing to do so may indicate a bug in the server.

I think it is useful when I change dependency from remote to local path without quiting and reopening nvim as I'm running flutter project, which LSP could resolve symbols from local path not from remote(~/.pub-cache/)

PR #172, opened 2022-07-29
feat: set dart comments options

### LINKED ISSUE #171: Multiple LSP servers attached to the same buffer
Issue is related to PR https://github.com/akinsho/flutter-tools.nvim/pull/163
and  revert of it:
https://github.com/akinsho/flutter-tools.nvim/commit/0bd239706f3a512b4619118a0f1e736976ff4f46

After revert I am getting the previous problem, that multiple LSP servers attach to the same buffer:
image
Looks like the original fix caused other problems, so I investigated more my config and found that issue disappears if I remove 
https://github.com/dart-lang/dart-vim-plugin from my config.

@akinsho, I see you mentioned this plugin in commit  https://github.com/akinsho/flutter-tools.nvim/commit/0417d4c45e732b5c96f4a42dbb562e5be0af81c1  - is this an intent to move some of `dart-vim-plugin` functionality to `flutter-tools.nvim`? What I miss if I remove `dart-vim-plugin`, is commentstring setup in https://github.com/dart-lang/dart-vim-plugin/blob/master/ftplugin/dart.vim
Or should both plugins work nicely together?

sidlatau:
Copied same configuration as is done there: https://github.com/dart-lang/dart-vim-plugin/blob/master/ftplugin/dart.vim#L16-L18

I also renamed the file from `lsp.lua` to `init.lua`, because now it contains code unrelated to LSP.

Closes #171

### DIFF
--- a/ftplugin/dart/init.lua
+++ b/ftplugin/dart/init.lua
@@ -1,4 +1,4 @@
 require("flutter-tools.lsp").attach()
 
-vim.api.nvim_buf_set_option(0, "comments", [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]])
-vim.api.nvim_buf_set_option(0, "commentstring", [[//%s]])
+vim.opt_local.comments = [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]]
+vim.opt_local.commentstring = [[//%s]]
--- a/ftplugin/dart/init.lua
+++ b/ftplugin/dart/init.lua
@@ -1,0 +1,4 @@
+require("flutter-tools.lsp").attach()
+
+vim.api.nvim_buf_set_option(0, "comments", [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]])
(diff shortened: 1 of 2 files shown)


ftplugin/dart/init.lua
```diff
@@ -0,0 +1,4 @@
+require("flutter-tools.lsp").attach()
+
+vim.api.nvim_buf_set_option(0, "comments", [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]])
```
akinsho, line 3: I generally prefer `vim.opt|o|opt_local` etc syntax as it looks "simpler" and more like a simple option setting than a function call
```suggestion
vim.opt_local.comments = [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]]
vim.opt_local.commentstring = [[//%s]]
```

akinsho (commented):
@sidlatau thanks for the PR. other than what I mentioned regarding using `vim.opt_local` this LGTM 👍🏿

ftplugin/dart/init.lua
sidlatau: Thanks for the comment, I was not aware of this syntax.

akinsho (approved):
LGTM 🚀 thanks for tweaking it

PR #173, opened 2022-07-29
feat: prevent writes to files in pub cache and FVM folder

sidlatau:


akinsho (changes requested):
@sidlatau I think all of this could go inside the ftplugin file since it's essentially equivalent to a file type autocommand

sidlatau:
> @sidlatau I think all of this could go inside the ftplugin file since it's essentially equivalent to a file type autocommand

Good point, moved code to ftplugin.

ftplugin/dart/init.lua
```diff
@@ -2,3 +2,11 @@ require("flutter-tools.lsp").attach()
 
 vim.opt_local.comments = [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]]
 vim.opt_local.commentstring = [[//%s]]
+
+local full_path = vim.fn.expand "%:p"
+-- Prevent writes to files in the pub cache and FVM folder.
+if string.find(full_path, ".pub-cache") or
+  string.find(full_path, [[Pub\Cache]]) or
+  string.find(full_path, "/fvm/versions/") then
```
akinsho:
```suggestion
if full_path:find(".pub-cache") or
  full_path:find([[Pub\Cache]]) or
  full_path:find("/fvm/versions/") then
```

ftplugin/dart/init.lua
```diff
@@ -2,3 +2,11 @@ require("flutter-tools.lsp").attach()
 
 vim.opt_local.comments = [[sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://]]
 vim.opt_local.commentstring = [[//%s]]
+
+local full_path = vim.fn.expand "%:p"
+-- Prevent writes to files in the pub cache and FVM folder.
+if string.find(full_path, ".pub-cache") or
+  string.find(full_path, [[Pub\Cache]]) or
```
akinsho: I'm not sure if any of these will need to be escaped as special characters in a lua pattern (thinking mostly about the `.` and the `-`). If you've checked them, and they work though, then 👍🏿

sidlatau: I escaped all the strings, visually looks consistent.

akinsho (approved):
LGTM 👍🏿

PR #179, opened 2022-08-26
prefer the dart binary located in dart_sdk rather than flutter_sdk

FlafyDev:
The `flutter` package in nix doesn't expose the dart binary in `<Flutter SDK>/bin/dart`. The binary is only found in the Dart SDK (`<Flutter SDK>/bin/cache/dart_sdk/bin/dart`).
This PR prefers the dart binary located in the Dart SDK rather than the Flutter SDK.

### Concers
This PR is ready to merge only if the dart binary is always located in the dart_sdk (which I'm pretty use is always the case, but idk).
If the dart_sdk path is found but there is no `<dart_sdk>/bin/dart` then the plugin would fail to get the dart binary. That's because the PR doesn't check if the dart binary exists, only if `paths.dart_sdk` doesn't exists then it checks `paths.flutter_sdk`.

### DIFF
--- a/lua/flutter-tools/executable.lua
+++ b/lua/flutter-tools/executable.lua
@@ -35,7 +35,7 @@ local function _dart_sdk_root(paths)
     return path.join(flutter_bin, dart_sdk)
   end
 
-  if utils.executable("dart") then return fn.resolve(fn.exepath("dart")) end
+  if utils.executable("dart") then return fn.resolve(path.dirname(path.dirname(fn.exepath("dart")))) end
 
   return ""
 end
--- a/lua/flutter-tools/executable.lua
+++ b/lua/flutter-tools/executable.lua
@@ -40,10 +40,11 @@ local function _dart_sdk_root(paths)
   return ""
 end
 
-local function _flutter_sdk_dart_bin(flutter_sdk)
-  -- retrieve the Dart binary from the Flutter SDK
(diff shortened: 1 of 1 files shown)


akinsho:
This seems potentially *very risky* and like it will introduce a bunch of problems for a plugin that is currently fairly stable. Can this not be solved in your config file by you just explicitly referencing where it should look for the binary. Tbh I haven't looked at all this logic in months so am defo rusty on the specifics.

FlafyDev:
> This seems potentially _very risky_ and like it will introduce a bunch of problems for a plugin that is currently fairly stable. Can this not be solved in your config file by you just explicitly referencing where it should look for the binary. Tbh I haven't looked at all this logic in months so am defo rusty on the specifics.

I could've just changed my Neovim config. But I wanted my Neovim to pick the Dart binary that is associated with the Flutter that is currently in PATH (because different projects have different flutter versions). Setting the location of the Dart binary in the Neovim config wont let me do it..

moonlitknight:
Just a thought, could you use the `flutter_lookup_cmd = nil, -- example "dirname $(which flutter)" or "asdf where flutter"` configuration option to do what you're looking for. This could reference a small script in which you can stat for files and return the PATH which you feel is most appropriate.

sidlatau:
Closing this as there is no progress in 2 years.

PR #183, opened 2022-09-08
Fixed trying to make upper on nil

FallenAngel97:
Hi. 
If the "opts" is empty table and level is being "nil", then ":upper()" makes exception. This should fix it


akinsho:
@FallenAngel97 thanks for raising this issue and letting me know ❤️ . I've just pushed a fix to the main branch, since I wanted to handle the default case differently. Can you let me know if this fixes the issue for you?

akinsho:
Gonna close this out since I believe it's been resolved

PR #188, opened 2022-10-20
feat: inlay hints

huylg:
Hi guys,

inlay_hints support for flutter and dart.

this pr just contains the implementation code. I will update README later :D

Demo:
image

Please take a look when you have any time. @akinsho

### DIFF
--- a/lua/flutter-tools.lua
+++ b/lua/flutter-tools.lua
@@ -111,6 +111,20 @@ local function setup_autocommands()
     })
   end
 
+  if require("flutter-tools.config").get("lsp").inlay_hints.enabled then
+    autocmd({ "BufEnter", "TextChanged", "InsertLeave" }, {
+      group = AUGROUP,
+      pattern = "*.dart",
+      callback = require("flutter-tools.lsp").inlay_hint,
+    })
+    autocmd({ "User" }, {
+      group = AUGROUP,
+      pattern = "FlutterToolsLspAnalysisComplete",
+      once = true,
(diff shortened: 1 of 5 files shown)


sidlatau:
I tested the code, it does not work yet in flutter `stable` branch, but in Flutter `master` hints are visible.

Not sure if they should be enabled by default until [anticonceal](https://github.com/neovim/neovim/pull/9496) is done. They do not add much value at the end of the line.

Example in `nvim`:
image

Example in VS Code:
image

huylg:
> I tested the code, it does not work yet in flutter `stable` branch, but in Flutter `master` hints are visible.
> 
> Not sure if they should be enabled by default until [anticonceal](https://github.com/neovim/neovim/pull/9496) is done. They do not add much value at the end of the line.
> 
> Example in `nvim`: image
> 
> Example in VS Code: image

thanks for review

I agree that inlay hints should be placed in exact characters that lsp return.
So I will disable inlay hints by default.

sidlatau:
Neovim nightly added support to inlay hints, so there is no need to add additional setup in this plugin anymore.
https://twitter.com/Neovim/status/1671046461071523843

```
inlay_hint({bufnr}, {enable}) vim.lsp.buf.inlay_hint()
Enable/disable/toggle inlay hints for a buffer
Parameters:
{bufnr} (integer) Buffer handle, or 0 for current
{enable} (boolean|nil) true/false to enable/disable, nil to toggle
```

PR #191, opened 2022-11-09
fix: do not attach LSP to buffer with diffview file URI

sidlatau:
I am using [diffview](https://github.com/sindrets/diffview.nvim) plugin, but when trying to open file history I am getting 4 error messages (see screenshot below). LSP server tries to attach to history files but fails because the path starts with `diffview://`. 

The simplest solution is to ignore these buffers. That is what I did in this PR. But not sure if this is the best solution - there could be other plugins that use other file prefixes, so in that case, we need to update the plugin again. One idea was not to attach LSP if the buffer is not modifiable, but that option is added later than attach event.

(We should also fix the problem, that there are 4 error messages instead of 2 and some of them are without any message)

image

### DIFF
--- a/lua/flutter-tools/lsp/init.lua
+++ b/lua/flutter-tools/lsp/init.lua
@@ -10,6 +10,7 @@ local fmt = string.format
 local FILETYPE = "dart"
 local SERVER_NAME = "dartls"
 local ROOT_PATTERNS = { ".git", "pubspec.yaml" }
+local INVALID_FILE_URIS = { "diffview://" }
 
 local M = {
   lsps = {},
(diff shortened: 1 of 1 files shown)


sidlatau:
I updated  PR to work in a more generic way: instead of checking for invalid buffer path prefixes it will check if the path has a URI prefix and if it does - is it a `file://` prefix. If not - LSP will not be attached. 
What is your opinion about this implementation, @akinsho?

akinsho:
@sidlatau that sounds good to me 👍🏿, always in favour of more generic solutions where possible

lua/flutter-tools/lsp/init.lua
```diff
@@ -229,6 +229,17 @@ local function legacy_server_init(bufnr, user_config)
   end)
 end
 
+--- Checks if buffer path is valid for attaching LSP
+local function is_valid_path(buffer_path)
+  local start_index, _, uri_prefix = buffer_path:find("^(%w+://).*")
+  if start_index and uri_prefix ~= "file://" then
```
akinsho, line 235: This could be simplified to 
```lua
return not start_index or uri_prefix == "file://"
```

Rather than return false, then true. The conditional might not be entirely right though, so please check.

akinsho (changes requested):
Looks good in general, just made one comment about maybe simplifying a conditional

akinsho (approved):
LGTM

PR #192, opened 2022-11-10
fix: do not show empty notifications

sidlatau:
As mentioned in https://github.com/akinsho/flutter-tools.nvim/pull/191 now the plugin sometimes shows 4 notifications, 2 of which are with an empty message. 

The concrete example when `lines` parameter is not an array of strings, but a table:
```lua
{                                                                                                               
  code = -32003,                                                                                                
  data = "diffview:///Users/ts/Documents/project/.git/c1795670f40/lib/app/launch_page.dart",  
  message = "URI was not a valid file:// URI",                                                                  
  <metatable> = {                                                                                               
    __tostring = <function 1>                                                                                   
  }                                                                                                             
}     
```


akinsho (approved):
LGTM thanks 🙏🏾

PR #198, opened 2022-12-28
fix(dap): read vmServiceUri for profiler_url

Thesmader:
Currently, when running via dap, the plugin is not able to detect the profiler_url.

While running with dap debugger, the plugin currently tries to set the profiler_url by trying to read `observatoryUri` from the body of 'event_dart.debuggerUris'.

But, the field `observatoryUri` is not available. The body instead has the field `vmServiceUri` with the websocket uri for the vm service. With this change, the plugin is now able to detect the profiler url and users can run the `FlutterCopyProfilerUrl` command to copy the devtools url


akinsho (approved):
LGTM 👍🏿 thanks for the fix

PR #204, opened 2023-02-01
feat: improve dev tools setup

sidlatau:
Fixed error handling when running :FlutterDevTools - `data` was expected to be table, but actually it is a string. 
New :FlutterDevToolsActivate command was added to simplify dev tools activation.


akinsho (approved):
LGTM, tbh I haven't really tested this out but it seems reasonable to me 🤞🏿 thanks

PR #216, opened 2023-03-05
fix(dap): use full path to flutter executable when running via DAP

sidlatau:
This fixes the problem with FVM - previously global flutter version was always used. After this change version is picked respecting FVM configuration.


akinsho (approved):
LGTM 👍🏾

PR #218, opened 2023-03-06
Make it so default setup can be copy pasted with error

tusqasi:
There are a few small things like missing commas and quotes which makes it a minor annoyance when copy-pasting.

This change will annihilate that annoyance.


akinsho:
@tusqasi is this meant to still be in draft?

PR #221, opened 2023-03-10
refactor(ui)!: replace (some) custom ui with `vim.ui.*`

akinsho:
When this plugin was first created almost none of the cool new APIs for making plugins existed. So I hand rolled a bunch of things like the menus, the tree layout etc. This stuff now accounts for like 40% of all the code in this repository if not more and there are now well implemented and tested libraries like nui.nvim which provide components for inputs, menus, trees, splits and other complex layouts. 

To simplify maintenance this PR removes all the custom logic for in favour of using nui.nvim for as much/all the custom UI.

### DIFF
--- a/.nvim.json
+++ b/.nvim.json
@@ -3,7 +3,8 @@
     "library": {
       "enabled": true,
       "plugins": [
-        "plenary.nvim"
+        "plenary.nvim",
+        "telescope.nvim"
       ]
     }
   }
--- a/README.md
+++ b/README.md
@@ -43,51 +43,46 @@ some basic setup might look like.
 
 - neovim 0.7.0+
 
-## Notices
-
-- This plugin no longer relies on dart code's debugger as flutter and dart now ship with a native debugger.
-  Versions of flutter before version 2.10.0 will still require the old dart code debugger.
-  Setup instructions for this can be found [here](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#dart).
-
-  Migration for existing users of the Dart code debugger on older flutter versions:
-
-  ```lua
-    local debugger_dir = path.join(fn.stdpath("cache"), "dart-code")
-    local debugger_path = path.join(debugger_dir, "out", "dist", "debug.js")
-
-    -- In config section for the debugger
-    debugger = {
-      enabled = true,
-      register.configurations = function()
-        local dap = require("dap")
-        dap.adapters.dart = {
-          type = "executable",
-          command = "node",
-          args = { debugger_path, "flutter" },
-        }
-        -- Other configuration herek
-      end,
-    },
-  ```
-
-- The `:FlutterOutline` command has been renamed to `:FlutterOutlineOpen`, and a `:FlutterOutlineToggle` has been added.
-
 ## Installation
 
 using `vim-plug`
 
 ```vim
 Plug 'nvim-lua/plenary.nvim'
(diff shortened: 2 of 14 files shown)


sidlatau:
What about using Telescope prompt for emulators dialog? 
image
I see it is already used in for commands dialog and has nicer alignment/paddings: 
image

fitrh:
Can this be opt-in?

akinsho:
@sidlatau I'd rather not have telescope become a hard requirement of this plugin because to be honest I personally am continually on the fence about if I'm going to continue to use it. Your point actually makes me realise that I can actually simplify things a bit using `vim.ui.select`, this way whatever the users preferred UI is is what will be used. I will still need nui for the complex layout like the outline tree.

Answering @fitrh using nui for the complex UI won't be optional since the trade off is maintaining a bunch of increasingly outdated UI code or improving the maintainability of this plugin by offloading this burden. I'm not sure what the opposition to adding this is but it will make things easier for me as the maintainer. I'll see about doing it in a way where any place it's used will just not work without it but things won't crash or error (graceful degradation)

fitrh:
As far as I understand, anything other than outline tree can use built-in
functionality like `vim.ui.*`

I may be biased as I'm more of *nvim-dap nvim-jdtls* guy

Besides outline tree, what other components require nui.nvim?

akinsho:
@fitrh whilst I appreciate the input I'm really not looking to box myself in on the basis of one person's preferences. Especially given the effort that myself and others have put into those project.

The goal of this project isn't to be hyperminimalist. Something which I think is evident from the existing feature set.

I don't intend to just add dependencies just because, but where and if I think necessary. If there's a `vim.ui.*` mechanism great I will go with that but I'm also not going to spend a bunch of my free time continuing to reinvent the wheel so people don't have to specify another plugin in their configs. 

It might be the case that there aren't many places where nui is needed, we will see as this PR is still in draft

akinsho:
@sidlatau can you have a look at the most recent version of this PR when you have a chance I've moved all the `menu|popup` usages to use `vim.ui.select` since it turns out *a lot* of the `ui` was collecting user input which can now be done with `ui.select` it completely hands over the ui to the persons plugin of choice so the rendering as you mentioned re. padding etc will depend on how they or their ui plugin has configured this.

I prefer this direction but I think a lot of new people trying out nvim dev coming from android studio and vscode will definitely stumble on `ui.select` if not configured so I'll probably add a link to `dressing.nvim` or something to the README. I've also added a kind and generally dressing let's you configure the UI per kind so I might also include a wiki page on how to customize things based on the kind of the `ui.select`

akinsho:
I'm likely going to boot rewriting the outline to `nui`'s tree because frankly it's more work than I have time for right now and I've been able to get rid of a lot of complexity by just switching to using `vim.ui.select`.

sidlatau:
@akinsho, thanks, I think this is a good approach, I actually get what I wanted: 
image

But the selection does not work - I select ios simulator, press ENTER and nothing happens.

akinsho:
@sidlatau thanks for testing it out. I should have mentioned I was still wrangling with the picker behaviour but was more looking to see if the UI was ok. I've fixed it now and refactored the telescope integration a bit so now the select menu if using telescope for selection via dressing should look the same as the plugin and the other menus should still work 🤞🏿

sidlatau:
Now selection works, LGTM 🚀

Issue #79, opened 2021-08-03
[Feature request] Project settings and custom telescope commands


ybbond:
hi, is there config to add default args for FlutterRun? Also, is there a way to add custom menu to aforementioned FlutterRun (with default args) for Telescope?

I currently use these configuration to achieve what I wanted:

in the config for `flutter-tools.nvim`
```lua
on_attach = function(client, bufnr)
  -- .....
  require('flutter-tools.utils').command("FlutterAgentStart", [[lua require('flutter-tools.commands').run_command('--flavor=staging --no-sound-null-safety')]])
  -- .....
end
```

I copied the whole `flutter-tools`'s file `menu.lua`, then added these lines:
```lua
      {
        id = "flutter-tools-run-agent",
        label = "Flutter tools: Start Agent",
        hint = "Start the Agent App project",
        command = function() require("flutter-tools.commands").run_command('--flavor=staging --no-sound-null-safety') end,
      },
```

then added my own telescope folder in:
```
lua
|- configs
|- telescope
    |- _extensions
        |- flutter.lua
```

Is there "better" way to achieve this? If there is currently none, this is a feature request 😄 

Thanks before!

akinsho:
Hi @ybbond,

Nope there currently isn't a way to do this, I discussed this a little bit over in #71. It isn't immediately clear how to do this easily tbh so I'm open to suggestions or PRs. I'm wary of adding in the wrong mechanism for this that would need to be deprecated later. For example vscode has a `launch.json` which a user can create per directory and so that informs what gets added to the run command etc. The problem with that is I'd have to create a custom per project config finding mechanism and avoid taken names like launch.json in case a user's project has one but they didn't intent to use it. Which feels like quite a bit of a scope creep into project settings territory for just one language specific plugin.

Alternatively it could be added to user's config but this then means it isn't project specific so you'd see run commands for other projects whilst in non-applicable projects. The solution could potentially be to namespace them under a particular directory in the config but then a user would have to add their project directory paths into their nvim config which is messy and maybe shares work information they don't want to share.

So TLDR, I'm not sure how to solve this problem.

akinsho:
[...]

PR #232, opened 2023-04-02
feat(config): add project configuration

akinsho:
Implements #79 
cc @sidlatau here's the branch I mentioned.

Currently the workflow is, a user creates an `exrc` file ie. a `.nvim.lua` or a `.nvimrc` and they add something like

```lua
require('flutter-tools').setup_project({
  device = "iPhone10",
  flavour = "Production",
})

-- Or
require('flutter-tools').setup_project({
	{
  		device = "iPhone10",
	    flavour = "Production",
		dart_define = {
			API_URL = "blah/thing.com",
		}
	},
	{
		device = "macOS",
		flavour = "Desktop",
	}
})
```

I've only implemented support for the device key atm but the others could be added eventually. The plan won't be to support everything immediately, but just to get a system in place for this PR.

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -98,13 +98,26 @@ function M.run_command(args)
   M.run({ args = args })
 end
 
----Run the flutter application
----@param opts {cli_args: string[]?, args: string[]?, device: Device?}
-function M.run(opts)
+---@param callback fun(project_config: config.ProjectConfig)
+local function select_project_config(callback)
+  local project_config = config.project
+  if #project_config < 2 then return callback(project_config[1]) end
+  vim.ui.select(project_config, {
+    prompt = "select a project configuration",
+    format_item = function(item) return vim.inspect(item) end,
+  }, function(selected)
+    if selected then callback(selected) end
+  end)
+end
+
+---@alias RunOpts {cli_args: string[]?, args: string[]?, device: Device?}
+
+---@param opts RunOpts
+---@param project_conf config.ProjectConfig?
+local function run(opts, project_conf)
   if M.is_running() then return ui.notify("Flutter is already running!") end
   opts = opts or {}
-  local project_config = config.project
-  local device = project_config.device or (opts.device and opts.device.id)
+  local device = project_conf and project_conf.device or (opts.device and opts.device.id)
   local cmd_args = opts.args
   local cli_args = opts.cli_args
   executable.get(function(paths)
@@ -122,6 +135,12 @@ function M.run(opts)
     runner = M.use_debugger_runner() and debugger_runner or job_runner
     runner:run(paths, args, lsp.get_lsp_root_dir(), on_run_data, on_run_exit)
   end)
+end
+
+---Run the flutter application
+---@param opts RunOpts
+function M.run(opts)
+  select_project_config(function(project_conf) run(opts, project_conf) end)
 end
 
 ---@param cmd string
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -99,18 +99,19 @@ function M.run_command(args)
 end
 
 ---Run the flutter application
----@param opts table
+---@param opts {cli_args: string[]?, args: string[]?, device: Device?}
 function M.run(opts)
   if M.is_running() then return ui.notify("Flutter is already running!") end
   opts = opts or {}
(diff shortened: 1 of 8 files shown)


sidlatau:
I want to add more details, about what problem I want to solve with project config: I have a project with two flavors and `dart-defines`(eg. DEV and PROD flavors). Now I have keymaps set up for each flavor in my dotfiles (`FlutterRun --flavor...`). But that is only related to the project, so it nice candidate for the project `.nvim.lua` configuration. So there are two configurations and I want an easy way to switch between them (maybe with the ability in lualine to see which configuration is currently selected).
Am I correct that in this PR the configuration will be set only on vim start?

akinsho:
@sidlatau yeah currently the idea was just picking one of the configs on startup but that's somewhat arbitrary since I don't have that use case. It can easily store both configs and then when you run start we can check if you have a project config and show the selection prompt at that stage? would that solve the runtime issue?

sidlatau:
Yes, that may work. I was also thinking about a separate command for setting configuration. But configuration should be set before running `FlutterRun`. So maybe there is no point in adding separate command if that is a part of the process of starting the `flutter` app.

akinsho:
@sidlatau yeah the configuration is IMO similar to a launch.json or something it's something that you can change over time but isn't intended to be some super dynamic thing that needs changing with a command presumably there are standard ways to run your project and those are fairly consistent and not things that are changing super often.

sidlatau:
Could support for `flavor` and `dart_define` be added to this PR too? I think these two adds most of the value plus then I would be able to help with testing this functionality with my project setup.

lua/flutter-tools/commands.lua
```diff
@@ -98,31 +100,61 @@ function M.run_command(args)
   M.run({ args = args })
 end
 
----Run the flutter application
----@param opts table
-function M.run(opts)
+---@param callback fun(project_config: flutter.ProjectConfig?)
+local function select_project_config(callback)
+  local project_config = config.project
+  if #project_config <= 1 then return callback(project_config[1]) end
```
sidlatau, line 106: I am thinking, that maybe we should also have "global config" too. Use case: I have two flavors, but by default, I want to start `flavor1`. That may even be true for device, it may also not be related to flavor.
When `default_config_name` is not provided then the selector will be shown as it is now. If provided - pick provided config. Example config:
```lua
{
  selected_config_name = 'Development',
  davice = 'android emulator',
  configs = {
    {
      flavor='dev',
      name='Development'
    },
    {
      flavor='prod',
      name='Production'
    },
  }
}
```

akinsho: @sidlatau I'd really rather keep this functionality simple (to start). I don't want this feature to become a complex config DSL on top of neovim, but something super straightforward/as dumb as possible. At least till it's clearer what is most valuable for most people. I want to merge a simple version and there presumably through reports over time it might become clearer what is just an inconvenience and what is vital.

In this case rather than this complex shape of structure you could just copy paste the device preferences, so it's quite simple to have the desired result.

I'd rather not be doing a bunch of logic to check for wholes in a table and then merges etc. I think you can just have both configs if you want one more than the other or don't want to be prompted about both then comment out one of just choose.

akinsho: Taking a bit of a look at dart code it seems to handle this stuff quite simply and dumb-ly
https://dartcode.org/docs/using-dart-define-in-flutter/

I'm not copying this exactly since it's not exactly what I want but I do like that it doesn't go down that sort of path and it's simple to understand at a glance I think

lua/flutter-tools/commands.lua
```diff
@@ -98,31 +100,61 @@ function M.run_command(args)
...
+  vim.ui.select(project_config, {
+    prompt = "select a project configuration",
+    format_item = function(item) return vim.inspect(item) end,
```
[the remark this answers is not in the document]
akinsho: That makes sense to me 👍🏾

sidlatau (approved):
The functionality works with my project configuration 👍🏻 
I think mentioning in README/docs is missing, and it will be LGTM

sidlatau:
One more thing: if the app is already started and I run `FlutterRun`, then I get the selector and then the info message:
``` 
Info  14:57:31 notify.info Flutter tools Flutter is already running!
```

Maybe it is possible to check if Flutter is already running before showing the selector?

Issue #233, opened 2023-04-04
Support file rename with imports update


sidlatau:
Imports are not automatically updated when a file is renamed. Dart LSP actually supports imports update via `workspace/willRenameFiles`, these changes need to be retrieved from the language server and applied after renaming.

This functionality in the future may be supported natively in NeoVim: https://github.com/neovim/neovim/issues/20784

But until that time I have implemented it here: https://github.com/sidlatau/dart-lsp-refactorings.nvim . I am using it without issues for several months. There are actually only two public functions: `require("dart-lsp-refactorings").rename()` that may be used instead of `vim.lsp.buf.rename()` and `require("dart-lsp-refactorings").on_rename_file` that may be used when renaming a file in file explorer (i.e neo-tree).

@akinsho, what is your opinion on bringing these or similar functions (with any needed code adjustments) to `flutter-tools` plugin? I see the benefits of having `dartls` related code in the same place. Also, more people may discover this functionality.
Or is it better to have it as a separate plugin? I understand that these functions are quite custom, plus it will add code that may require additional maintenance.

Related to: https://github.com/akinsho/flutter-tools.nvim/issues/39

akinsho:
@sidlatau sorry been a bit busy, I'm definitely open to adding this. FWIW although I personally don't have the time or bandwidth to try to emulate everything (for flutter) in Android Studio or VSCode I think the general goal of this plugin is to make developing flutter/dart apps on neovim as nice as possible and take inspiration from those tools as the gold standards for flutter dev.

Anyway all that being said the only reason I had not added this is that I had hoped that a native version would have been merged by now. Another reason is that dart LS added in some options around this.
https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md#client-workspace-configuration

i.e. so we already get renaming a file for free if the class matches the file, the other one that is supported there is `dart.updateImportsOnRename` which I guess this feature would add support for until there is a neovim native solution

Looking through your project briefly I think the language server is probably doing some of this work that that plugin does automatically and the only change needed here to make all this work is to add support for `workspace/willRenameFiles` once we have signaled that support to the server and have handlers for it I think it should work.

I wonder if we would actually need separate commands since I'd love to do this in such a way that it is easy to deprecate once neovim adds support for this

sidlatau:
[...]

PR #234, opened 2023-04-06
feat: add FlutterRename command

sidlatau:
This command adds support for `updateImportsOnRename` setting (https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md#client-workspace-configuration). 

It is useful when the file is renamed after a class rename (when enabled `renameFilesWithClasses` setting). 
It will fall back to `vim.lsp.buf.rename()` if no `dartls` client is found. 
If the client is found, the file is the same as the class name, and the class is being renamed,  it will request for changes via `workspace/willRenameFiles`. Results of `workspace/willRenameFiles` will be applied after the class rename.

To test it try to rename the class that is in the file with the same file name. 

- with `vim.lsp.buf.rename()`: class renamed, file renamed, usage of class renamed, but imports are not updated.
- with `FlutterRename`: class renamed, file renamed, usage and imports are updated.

### DIFF
--- a/lua/flutter-tools/lsp/rename.lua
+++ b/lua/flutter-tools/lsp/rename.lua
@@ -1,9 +1,11 @@
 local M = {}
 
-local api = vim.api
-local util = vim.lsp.util
 local lazy = require("flutter-tools.lazy")
 local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
+local lsp_utils = lazy.require("flutter-tools.lsp.utils") ---@module "flutter-tools.lsp.utils"
+
+local api = vim.api
+local util = vim.lsp.util
 
 --- Computes a filename for a given class name (convert from PascalCase to  snake_case).
 local function file_name_for_class_name(class_name)
@@ -38,11 +40,7 @@ end
 function M.rename(new_name, options)
   options = options or {}
   local bufnr = options.bufnr or api.nvim_get_current_buf()
-  local clients = vim.lsp.get_active_clients({
-    bufnr = bufnr,
-    name = "dartls",
-  })
-  local client = clients[1]
+  local client = lsp_utils.get_dartls_client(bufnr)
   if not client then
     -- Fallback to default rename function if language server is not dartls
     vim.lsp.buf.rename(new_name, options)
--- a/lua/flutter-tools/lsp/rename.lua
+++ b/lua/flutter-tools/lsp/rename.lua
@@ -1,11 +1,14 @@
 local M = {}
 
 local lazy = require("flutter-tools.lazy")
-local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
 local lsp_utils = lazy.require("flutter-tools.lsp.utils") ---@module "flutter-tools.lsp.utils"
+local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
+local ui = lazy.require("flutter-tools.ui") ---@module "flutter-tools.ui"
 
 local api = vim.api
 local util = vim.lsp.util
+local lsp = vim.lsp
+local fn = vim.fn
 
 --- Computes a filename for a given class name (convert from PascalCase to  snake_case).
 local function file_name_for_class_name(class_name)
@@ -18,16 +21,16 @@ local function file_name_for_class_name(class_name)
 end
 
 local function will_rename_files(old_name, new_name, callback)
-  local params = vim.lsp.util.make_position_params()
+  local params = lsp.util.make_position_params()
   if not new_name then return end
   local file_change = {
     newUri = vim.uri_from_fname(new_name),
     oldUri = vim.uri_from_fname(old_name),
   }
   params.files = { file_change }
-  vim.lsp.buf_request(0, "workspace/willRenameFiles", params, function(err, result)
+  lsp.buf_request(0, "workspace/willRenameFiles", params, function(err, result)
     if err then
-      vim.notify(err.message or "Error on getting lsp rename results!")
+      ui.notify(err.message or "Error on getting lsp rename results!", ui.ERROR)
       return
     end
     callback(result)
@@ -43,15 +46,14 @@ function M.rename(new_name, options)
   local client = lsp_utils.get_dartls_client(bufnr)
   if not client then
     -- Fallback to default rename function if language server is not dartls
-    vim.lsp.buf.rename(new_name, options)
-    return
+    return lsp.buf.rename(new_name, options)
   end
 
   local win = api.nvim_get_current_win()
 
   -- Compute early to account for cursor movements after going async
-  local cword = vim.fn.expand("<cword>")
-  local actual_file_name = vim.fn.expand("%:t")
+  local cword = fn.expand("<cword>")
+  local actual_file_name = fn.expand("%:t")
   local old_computed_filename = file_name_for_class_name(cword)
   local is_file_rename = old_computed_filename == actual_file_name
 
@@ -69,23 +71,22 @@ function M.rename(new_name, options)
   local function rename(name, will_rename_files_result)
     local params = util.make_position_params(win, client.offset_encoding)
     params.newName = name
-    local handler = client.handlers["textDocument/rename"]
-      or vim.lsp.handlers["textDocument/rename"]
+    local handler = client.handlers["textDocument/rename"] or lsp.handlers["textDocument/rename"]
     client.request("textDocument/rename", params, function(...)
       handler(...)
       if will_rename_files_result then
         -- the `will_rename_files_result` contains all the places we need to update imports
         -- so we apply those edits.
-        vim.lsp.util.apply_workspace_edit(will_rename_files_result, client.offset_encoding)
+        lsp.util.apply_workspace_edit(will_rename_files_result, client.offset_encoding)
       end
     end, bufnr)
   end
 
   local function rename_fix_imports(name)
     if is_file_rename then
-      local old_file_path = vim.fn.expand("%:p")
+      local old_file_path = fn.expand("%:p")
       local new_filename = file_name_for_class_name(name)
-      local actual_file_head = vim.fn.expand("%:p:h")
+      local actual_file_head = fn.expand("%:p:h")
       local new_file_path = path.join(actual_file_head, new_filename)
       will_rename_files(old_file_path, new_file_path, function(result) rename(name, result) end)
     else
@@ -99,7 +100,7 @@ function M.rename(new_name, options)
       if err or result == nil then
         local msg = err and ("Error on prepareRename: " .. (err.message or ""))
           or "Nothing to rename"
-        vim.notify(msg, vim.log.levels.INFO)
+        ui.notify(msg, ui.INFO)
         return
       end
 
--- a/lua/flutter-tools/lsp/rename.lua
+++ b/lua/flutter-tools/lsp/rename.lua
@@ -72,7 +72,7 @@ function M.rename(new_name, options)
     local params = util.make_position_params(win, client.offset_encoding)
     params.newName = name
     local handler = client.handlers["textDocument/rename"]
-        or vim.lsp.handlers["textDocument/rename"]
+      or vim.lsp.handlers["textDocument/rename"]
     client.request("textDocument/rename", params, function(...)
       handler(...)
       if will_rename_files_result then
@@ -100,7 +100,7 @@ function M.rename(new_name, options)
     client.request("textDocument/prepareRename", params, function(err, result)
       if err or result == nil then
         local msg = err and ("Error on prepareRename: " .. (err.message or ""))
-            or "Nothing to rename"
+          or "Nothing to rename"
         vim.notify(msg, vim.log.levels.INFO)
         return
       end
--- a/lua/flutter-tools/lsp/rename.lua
+++ b/lua/flutter-tools/lsp/rename.lua
@@ -1,0 +1,149 @@
+local M = {}
+
+local api = vim.api
+local util = vim.lsp.util
+local lazy = require("flutter-tools.lazy")
+local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
+
+--- Computes a filename for a given class name (convert from PascalCase to  snake_case).
+local function file_name_for_class_name(class_name)
+  local starts_uppercase = class_name:find("^%u")
+  if not starts_uppercase then return nil end
+  local file_name = class_name:gsub("(%u)", "_%1"):lower()
+  -- Removes first underscore
+  file_name = file_name:sub(2)
+  return file_name .. ".dart"
+end
+
+local function will_rename_files(old_name, new_name, callback)
+  local params = vim.lsp.util.make_position_params()
+  if not new_name then return end
+  local file_change = {
+    newUri = vim.uri_from_fname(new_name),
+    oldUri = vim.uri_from_fname(old_name),
+  }
+  params.files = { file_change }
+  vim.lsp.buf_request(0, "workspace/willRenameFiles", params, function(err, result)
+    if err then
+      vim.notify(err.message or "Error on getting lsp rename results!")
+      return
+    end
+    callback(result)
+  end)
+end
+
+--- Call this function when you want rename class or anything else.
+--- If file will be renamed too, this function will update imports.
+--- Function has same signature as `vim.lsp.buf.rename()` function and can be used instead of it.
+function M.rename(new_name, options)
+  options = options or {}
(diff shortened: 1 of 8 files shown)


lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+  end)
+end
+
+--- Call this function when you want rename class or anything else.
+--- If file will be renamed too, this function will update imports.
+--- Function has same signature as `vim.lsp.buf.rename()` function and can be used instead of it.
+function M.rename(new_name, options)
+  options = options or {}
+  local bufnr = options.bufnr or api.nvim_get_current_buf()
+  local clients = vim.lsp.get_active_clients({
```
akinsho, line 41: I think there is already a function to get `dartls` in the `lsp` module this means we can reuse that logic here and ensure it's done the same way everywhere

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+        -- the `will_rename_files_result` contains all the places we need to update imports
+        -- so we apply those edits.
+        vim.lsp.util.apply_workspace_edit(will_rename_files_result, client.offset_encoding)
+      end
+    end, bufnr)
+  end
+
+  local function rename_fix_imports(name)
+    if is_file_rename then
+      local old_file_path = vim.fn.expand("%:p")
```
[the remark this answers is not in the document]
akinsho: Also could you use an api function for this something like `api.nvim_get_buf_name(0)` and for filename transformations you could use `vim.fs.dirname` etc.? unless it's actually complex and you need `filename-modifiers`

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+    client.request("textDocument/prepareRename", params, function(err, result)
+      if err or result == nil then
+        local msg = err and ("Error on prepareRename: " .. (err.message or ""))
+          or "Nothing to rename"
+        vim.notify(msg, vim.log.levels.INFO)
+        return
+      end
+
+      if new_name then
+        rename_fix_imports(new_name)
```
akinsho, line 109: I'd combine the return here also

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+        return
+      end
+
+      local prompt_opts = {
```
akinsho, line 113: Think this can be one line 🤷🏾‍♂️

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+      if result.placeholder then
+        prompt_opts.default = result.placeholder
+      elseif result.start then
+        prompt_opts.default = get_text_at_range(result, client.offset_encoding)
+      elseif result.range then
+        prompt_opts.default = get_text_at_range(result.range, client.offset_encoding)
+      else
+        prompt_opts.default = cword
+      end
+      vim.ui.input(prompt_opts, function(input)
```
akinsho, line 126: I think maybe `vim.ui.input` should also be in `ui` since if there's metadata to pass in the future or we want to swap out the implementation to something else it won't then require hunting all these naked calls down.

akinsho (commented):
Just had a quick look, not a super detailed review atm. Looks good generally, just some stylistic comments. Also one somewhat big comment which is mainly I wonder if a lot of this logic isn't already handled by the lsp stuff like calculating which files to update etc. also I wonder if `vim.lsp.util.rename` can't also be used here

akinsho:
will fix #233

lua/flutter-tools/lsp/rename.lua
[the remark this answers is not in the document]
sidlatau: Added aliases. For the second comment I am not really sure how to use that `vim.fs.dirname`. Will that simplify the code?

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,149 @@
...
+      handler(...)
+      if will_rename_files_result then
+        -- the `will_rename_files_result` contains all the places we need to update imports
+        -- so we apply those edits.
+        vim.lsp.util.apply_workspace_edit(will_rename_files_result, client.offset_encoding)
+      end
+    end, bufnr)
+  end
+
+  local function rename_fix_imports(name)
```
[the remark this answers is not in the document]
sidlatau: Looks like it does not. Imports update changes need to be fetched before rename and updated after rename.
As I mentioned in PR summary that could be tested in this way:
> To test it try to rename the class that is in the file with the same file name.

> - with vim.lsp.buf.rename(): class renamed, file renamed, usage of class renamed, but imports are not updated.
> - with FlutterRename: class renamed, file renamed, usage and imports are updated.

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+
+local api = vim.api
+local util = vim.lsp.util
+local lsp = vim.lsp
+local fn = vim.fn
+
+--- Computes a filename for a given class name (convert from PascalCase to  snake_case).
+local function file_name_for_class_name(class_name)
+  local starts_uppercase = class_name:find("^%u")
+  if not starts_uppercase then return nil end
```
akinsho, line 16:
```suggestion
  if not starts_uppercase then return end
```
No point in the explicit `nil` IMO

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+local lsp_utils = lazy.require("flutter-tools.lsp.utils") ---@module "flutter-tools.lsp.utils"
+local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
+local ui = lazy.require("flutter-tools.ui") ---@module "flutter-tools.ui"
+
+local api = vim.api
+local util = vim.lsp.util
+local lsp = vim.lsp
+local fn = vim.fn
+
+--- Computes a filename for a given class name (convert from PascalCase to  snake_case).
```
akinsho, line 13:
```suggestion
--- Computes a filename for a given class name (convert from PascalCase to  snake_case).
---@param class_name string
---@return string?
```

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+local function file_name_for_class_name(class_name)
+  local starts_uppercase = class_name:find("^%u")
+  if not starts_uppercase then return nil end
+  local file_name = class_name:gsub("(%u)", "_%1"):lower()
+  -- Removes first underscore
+  file_name = file_name:sub(2)
+  return file_name .. ".dart"
+end
+
+local function will_rename_files(old_name, new_name, callback)
```
akinsho, line 23: Also recommend a type annotation here. `lua_ls` can use type annotations the more detailed they are to actually ensure things like `nil` checks and correct number of arguments etc.

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+  local params = lsp.util.make_position_params()
+  if not new_name then return end
+  local file_change = {
+    newUri = vim.uri_from_fname(new_name),
+    oldUri = vim.uri_from_fname(old_name),
+  }
+  params.files = { file_change }
+  lsp.buf_request(0, "workspace/willRenameFiles", params, function(err, result)
+    if err then
+      ui.notify(err.message or "Error on getting lsp rename results!", ui.ERROR)
```
akinsho, line 33:
```suggestion
return ui.notify(err.message or "Error on getting lsp rename results!", ui.ERROR)
```

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+    -- Fallback to default rename function if language server is not dartls
+    return lsp.buf.rename(new_name, options)
+  end
+
+  local win = api.nvim_get_current_win()
+
+  -- Compute early to account for cursor movements after going async
+  local cword = fn.expand("<cword>")
+  local actual_file_name = fn.expand("%:t")
+  local old_computed_filename = file_name_for_class_name(cword)
```
akinsho, line 57: I think `prev_*` and `curr_*` are *to me* more clear than `old_*` and `actual_*` I can't tell if old is what the thing used to be or if it's what it is currently etc.

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+  if not client then
+    -- Fallback to default rename function if language server is not dartls
+    return lsp.buf.rename(new_name, options)
+  end
+
+  local win = api.nvim_get_current_win()
+
+  -- Compute early to account for cursor movements after going async
+  local cword = fn.expand("<cword>")
+  local actual_file_name = fn.expand("%:t")
```
akinsho, line 56: I use `filename-modifiers` a lot but I worry that they make contributing to lua projects harder a there's a lot of magic in them. Someone has to know that `%` is the current file and that `:t` will take the tail of the file. Although more verbose which is something I usually avoid, I'm trying to introduce.

`(vim.)fs.basename(api.nvim_buf_get_name(bufnr))` as an alternative to filename modifiers. I think you can read this as a none vim person or newer person and know what this probably means.

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+  local old_computed_filename = file_name_for_class_name(cword)
+  local is_file_rename = old_computed_filename == actual_file_name
+
+  local function get_text_at_range(range, offset_encoding)
+    return api.nvim_buf_get_text(
+      bufnr,
+      range.start.line,
+      util._get_line_byte_from_position(bufnr, range.start, offset_encoding),
+      range["end"].line,
+      util._get_line_byte_from_position(bufnr, range["end"], offset_encoding),
```
akinsho, line 66: Are these private neovim methods? if so we should document somewhere clearly that these can break in the future if deprecated or renamed

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+      {}
+    )[1]
+  end
+
+  local function rename(name, will_rename_files_result)
```
akinsho, line 71: I think we can just call the value result rather than `will_rename_files_result` and then you can add extra information in a type annotation e.g.
```lua
---@param name string the name of the thing
---@param result {thing: string} the result from the call to will rename
```

lua/flutter-tools/lsp/rename.lua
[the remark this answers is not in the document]
akinsho: Fair enough I'm surprised that each consumer is expected to implement this or is the logic here largely what would go inside of a neovim client implementation of will rename ? (OOI)

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+      if will_rename_files_result then
+        -- `will_rename_files_result` contains all the places we need to update imports, so we apply those edits.
+        lsp.util.apply_workspace_edit(will_rename_files_result, client.offset_encoding)
+      end
+    end, bufnr)
+  end
+
+  local function rename_fix_imports(name)
+    if is_file_rename then
+      local old_file_path = fn.expand("%:p")
```
akinsho, line 86: Similar comment about old here, it's probably just me but old makes me ask questions like, is something time related happening or something more subtle whereas `prev_*` just makes me think something got re-assigned.

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+      local new_filename = file_name_for_class_name(name)
+      local actual_file_head = fn.expand("%:p:h")
```
akinsho, line 88: similarly I think this can be
```lua
fs.basename(api.nvim_buf_get_name())
```
I appreciate these are more verbose which is normally against what I prefer but I think it's best to save modifiers for real complex path modification stuff that would require a complex lua function to replicate

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  if client.supports_method("textDocument/prepareRename") then
+    local params = util.make_position_params(win, client.offset_encoding)
+    client.request("textDocument/prepareRename", params, function(err, result)
+      if err or result == nil then
+        local msg = err and ("Error on prepareRename: " .. (err.message or ""))
+          or "Nothing to rename"
+        ui.notify(msg, ui.INFO)
+        return
```
akinsho, line 103:
```suggestion
return ui.notify(msg, lvl)
```

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+      will_rename_files(old_file_path, new_file_path, function(result) rename(name, result) end)
+    else
+      rename(name)
+    end
+  end
+
+  if client.supports_method("textDocument/prepareRename") then
+    local params = util.make_position_params(win, client.offset_encoding)
+    client.request("textDocument/prepareRename", params, function(err, result)
+      if err or result == nil then
```
akinsho, line 99:
```suggestion
      if err or not result then
```

I only compare against `false` or `nil` if the value is subtle if all falsy should be treated the same then it's clearer to just do `not value` or `if value`

whereas if you wanted to differentiate between `value == nil then one_thing` and `value == false then another_thing` then explicit checks make sense

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+        local msg = err and ("Error on prepareRename: " .. (err.message or ""))
```
akinsho, line 100: If the type of `err.message` is a `string` then you don't need this `or` you can just do
```lua
('Error in prepareRename: %s'):format(err.message)
```

lua/flutter-tools/lsp/rename.lua
```diff
@@ -0,0 +1,142 @@
...
+      ui.input(prompt_opts, function(input)
+        if not input or #input == 0 then return end
+        rename_fix_imports(input)
+      end)
+    end, bufnr)
+  else
+    assert(client.supports_method("textDocument/rename"), "Client must support textDocument/rename")
+    if new_name then
+      rename_fix_imports(new_name)
+      return
```
akinsho, line 128:
```suggestion
      return rename_fix_imports(new_name)
```

akinsho (commented):
@sidlatau thanks for factoring out the `ui.input` calls and the `lsp.get_client` stuff makes things much cleaner. The remaining stuff I've flagged are almost entirely stylistic. Things like
* Adding type annotations to make the code easier to use with lua ls
* preferring `old` -> `prev`, `actual` -> `current`
* `expand("magic")` -> `vim.fs.(base|dir)name(api.nvim_buf_get_name())`

lua/flutter-tools/lsp/rename.lua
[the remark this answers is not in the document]
akinsho: @sidlatau the simplification is more aimed at new contributors to the code that might not be super well versed in vim magic since things like `%:t:t:r` mean nothing to anyone who hasn't been doing vim stuff for a while. This is one area where I trade terseness for readability. Re. using vim.fs if you do `:h vim.fs.dirname` etc. those methods are all documented and I think are pretty straightforward in their usage

lua/flutter-tools/lsp/rename.lua
sidlatau: Yes, I added a direct link to the current snapshot of implementation in case it will change in the future.
As I mentioned in the function header comment, it is a modified version of `vim.lsp.buf.rename`, the code is just adjusted to work together with  `workspace/willRenameFiles` request.

lua/flutter-tools/lsp/rename.lua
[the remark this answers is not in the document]
sidlatau: As I understand it, this is done so to make `renameFilesWithClasses` setting work. This is not a standard LSP setting. So manual coordination is required to work together with `updateImportsOnRename` setting. Plugin informs LSP that the file will be renamed (by providing oldUri, newUri server calculates changes). Then rename of the class is performed. Then `renameFilesWithClasses` logic is applied. And after that `willRenameFiles` result should be applied.

After writing all this I just realized, that all this rename logic is relevant only if `renameFilesWithClasses = "always",` I will add an additional check for it :)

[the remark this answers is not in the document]
sidlatau: Updated code b2dbaefd4755522b928ca27724757c3124f0c4ed

akinsho (approved):
LGTM 👍🏾

Issue #252, opened 2023-05-03
[BUG] Autoscroll spams notify when there are too many logs


martyfuhry:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

As we all have experienced, sometimes the Flutter dev log (especially in Android) can be spammed with messages. When this happens, [the autoscroll behavior starts spamming vim.notify](https://github.com/akinsho/flutter-tools.nvim/blob/b3854c50aef753ed52bf15517af7eaa82b46a339/lua/flutter-tools/log.lua#L65), which eventually makes my IDE unusable. I'm unable to get past the endless notifications from the autoscroll to even kill Flutter.

**50,000+ lines of ADB Log Spam**

![image](https://user-images.githubusercontent.com/100457/236020990-4179e675-1cd2-44d8-8399-c22538acd889.png)

**`vim.notify` Spam**

![image](https://user-images.githubusercontent.com/100457/236017595-5555b2d9-834b-4294-ab04-ba64d7cf166d.png)

### Expected Behavior

I would like to be able to suppress all of those notifications. I get it, ADB is spamming the console output. But it shouldn't hang up my entire IDE.

I don't know that the autoscroll should even ping `vim.notify`. Maybe suppress this message entirely. Maybe only log it to the dev logs? Maybe allow someone to suppress this particular message through configuration settings.

### Steps To Reproduce

**Config**

```lua
  flutter_path = "/usr/local/lib/fvm/default",
  fvm = true,
  lsp = {
    on_attach = on_attach,
  },
```

**Example Code?**
```dart

while (true) {
  await Future.delayed(Duration(milliseconds: 1));
  print('SPAM');
}
```

1. Run some code which spams output to ADB. Maybe something like the above? For me it happens with Flutter's Google Maps library (I get : ` BufferQueueProducer ... BufferQueue has been abandoned` spam).
2. Run Flutter using `flutter-tools`.
3. Keep the `__FLUTTER_DEV_LOG__` buffer open with autoscroll enabled.
4. Move your cursor to some line on the buffer.
5. Get spammed with notifications and the IDE becomes unusable.

### Environment

```Markdown
- OS:Pop!_OS
- Flutter version: 3.7.9
- Is flutter in $PATH: FVM is
- neovim version: v0.10.0-dev
```

### Anything else?

I love this project. Also, I'm using NVChad if that's relevant? If there's anything else I can help with or if you need additional info, please let me know!

[...]

PR #253, opened 2023-05-04
Notify once for autoscroll

martyfuhry:
This seems to fix my issue with #252 

image

I see this notification one time instead of being bombarded! Thank you for the help.


akinsho:
@martyfuhry thanks for the PR, I decided to add an option in to the wrapper function this plugin uses and couldn't push to this branch. But my change is effectively the same as yours. Hopefully that works

martyfuhry:
I'm sure this works better than mine! Thank you for making the change.

PR #255, opened 2023-05-06
chore: use `dart language-server --protocol=lsp`

Decodetalkers:
use dart language-server --protocol=lsp to start lsp

it should not find snapshot manually

### DIFF
--- a/lua/flutter-tools/lsp/init.lua
+++ b/lua/flutter-tools/lsp/init.lua
@@ -16,10 +16,6 @@ local M = {
   lsps = {},
 }
 
-local function analysis_server_snapshot_path(sdk_path)
-  return path.join(sdk_path, "bin", "snapshots", "analysis_server.dart.snapshot")
-end
-
 ---Merge a set of default configurations with a user's own settings
 --- NOTE: a user can specify a function in which case this will be used
 ---to determine how to merge the defaults with a user's config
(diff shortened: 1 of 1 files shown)


akinsho:
@Decodetalkers thanks for this PR, it looks like a great addition. Definitely good to get out of the snapshot finding business. One question, do you know what version of Flutter this was added in i.e. who will this break compatibility for. If it's over one major version ago then I'm fine to add this change but if it's just released then we'd need to maintain backwards compatibility for people who can't easily change their flutter version themselves

Decodetalkers:
> @Decodetalkers thanks for this PR, it looks like a great addition. Definitely good to get out of the snapshot finding business. One question, do you know what version of Flutter this was added in i.e. who will this break compatibility for. If it's over one major version ago then I'm fine to add this change but if it's just released then we'd need to maintain backwards compatibility for people who can't easily change their flutter version themselves

I don't know.. but in nvim-lspconfig.nvim , it is used as this way.. I think we should keep the same as nvim-lspconfig.. and if the user cannot , they should define the command them self.

I have find the readme. seems form 1.2.1,  dart has already had the option of `language-server`, I think user can check `dart --help`, if they have the option of `language-server`, if none ,they should pass the command them self
https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/tool/lsp_spec/README.md

akinsho:
Ok well v1.2.1 is ancient so this is good 2 go for me

PR #259, opened 2023-05-11
Support nested dart sdk

fb87:
Some system like NixOS, the dart is nested within `flutter/bin` which leads to the error where dart will not be found as current implementation. This patch adds the `flutter/bin/cache/dart-sdk/bin` to the search paths along with `flutter/bin`. This is short term solution, for long term, IMO we can search for the executable dart within `flutter/bin`.

### DIFF
--- a/lua/flutter-tools/executable.lua
+++ b/lua/flutter-tools/executable.lua
@@ -44,7 +44,18 @@ end
 local function _flutter_sdk_dart_bin(flutter_sdk)
   -- retrieve the Dart binary from the Flutter SDK
   local binary_name = path.is_windows and "dart.bat" or "dart"
-  return path.join(flutter_sdk, "bin", binary_name)
+  local dart_search_paths = {
+    path.join(flutter_sdk, "bin", binary_name),
+    path.join(flutter_sdk, "bin", "cache", "dart-sdk", "bin", binary_name)
+  }
+
+  for _, _path in ipairs(dart_search_paths) do
+    if path.exists(_path) then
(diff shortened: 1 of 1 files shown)


akinsho:
Hi @fb87,

Is this problem specific to NixOS? there was also an attempt in #179 which frankly I didn't like the look of but this looks more robust than that. Generally I really want to avoid having logic for specific OSes, I really didn't but should have realised what a nasty rabbit hole I was getting into handling any of the path stuff for users and should frankly have just said something like make sure the binaries are in the $PATH for vim or use the full path. Anyway ship has sailed now.

I think rather than just adding another path to search I'd rather first see if the full path passed to the config can't work for Nix or anyone else who's path is more custom than what is already handled and if not see if theres some way to just expose something that the user can take control of and do with what they may rather than continue down the path of more and more edge cases for different OSes.

fb87:
The binaries (`dart`, `flutter`) are in the `PATH` already, the main issue is `_flutter_sdk_dart_bin(...)` returns the path of `flutter/bin/dart`. We might introduce the other options which allow user to specify where the `dart` is, such as `dart_sdk_path = "<full/path/if/needed>"` or `dart_lookup_cmd`.

akinsho:
> IMO, we can introduce the other options which allows user to specify where the dart is, such as dart_sdk_path = "<full/path/if/needed>" or dart_lookup_cmd.

I think this would be good since hopefully the combination will close out all future requests for custom locations since users will now just be able to set it themselves

Issue #224, opened 2023-03-16
Can't use commands


Neeraj029:
Hi!

I just installed flutter-tools using lazy.nvim, lsp for dart works fine but Flutter commands aren't showing up
I just used the code given for lazy.nvim on the install section.
![image](https://user-images.githubusercontent.com/66945520/225533252-5491743a-c856-4cc9-bcff-eda5f3d49ac3.png)

Any help would be appreciated!

akinsho:
Hi @Neeraj029,

so you're missing some crucial information from this issue i.e. how do you configure this plugin did you call setup at some point in the plugin configuration? there isn't an example of a minimal configuration to show what hasn't worked and under what circumstance.

I suggest carefully going through the README about how this plugin is started and making sure you've done that in a way that works with lazy.nvim. As a lazy.nvim user myself I know that this plugin will work with it so this issue is a configuration issue.

Neeraj029:
Hi @akinsho 
![image](https://user-images.githubusercontent.com/66945520/225573473-189c792d-2b3a-4d8b-ab79-529af8cacebc.png)
I'm doing this

akinsho:
@Neeraj029 in general it's better to paste code rather than a screenshot of code so I can easily copy and paste what you are showing me. In this case it's clear from your setup that you are not explicitly calling setup for this plugin. In general I strongly advice you read in detail the README of this plugin and of the configuration framework you are using as this is an important step that you are missing and maybe don't understand.

you need something like
```lua
config = function()
	require('flutter-tools').setup()
end
```

I think you should be able to handle the rest by carefully reading the bits of documentation for these tools and this doesn't need to be an ongoing issue on this repo

PR #262, opened 2023-05-31
Add lazy.nvim install with integrated setup()

jidn:
Change documentation to show the lazy.nvim install that includes the flutter-tools setup. Eliminates questions like issue #224.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ using `lazy.nvim`
         'nvim-lua/plenary.nvim',
         'stevearc/dressing.nvim', -- optional for vim.ui.select
     },
-    opts = {}   -- same as calling `require('flutter-tools').setup({})`
+    opts = true,
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ using `lazy.nvim`
         'nvim-lua/plenary.nvim',
         'stevearc/dressing.nvim', -- optional for vim.ui.select
     },
-    opts = true,
+    config = true,
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ using `lazy.nvim`
         'nvim-lua/plenary.nvim',
         'stevearc/dressing.nvim', -- optional for vim.ui.select
     },
+    opts = {}   -- same as calling `require('flutter-tools').setup({})`
 }
 ```
 
--- a/doc/flutter-tools.txt
+++ b/doc/flutter-tools.txt
@@ -106,6 +106,7 @@ using `lazy.nvim`
             'nvim-lua/plenary.nvim',
             'stevearc/dressing.nvim', -- optional for vim.ui.select
         },
+        opts = {}   -- same as calling `require('flutter-tools').setup({})`
     }
 <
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -75,6 +75,7 @@ using `lazy.nvim`
         'nvim-lua/plenary.nvim',
         'stevearc/dressing.nvim', -- optional for vim.ui.select
     },
+    opts = {}   -- same as calling `require('flutter-tools').setup({})`
```
akinsho, line 78: @jidn you can also just use `config = true` which I think is a bit better than having to pass an empty table and have that comment

doc/flutter-tools.txt
```diff
@@ -106,6 +106,7 @@ using `lazy.nvim`
             'nvim-lua/plenary.nvim',
             'stevearc/dressing.nvim', -- optional for vim.ui.select
         },
+        opts = {}   -- same as calling `require('flutter-tools').setup({})`
```
akinsho, line 109: The help docs are generated from the README so there's no need to update it manually

jidn: You are right, it is cleaner.   I don't know if `config = true` is any more performant. Although, it does provide an obvious way to add setup configuration, but I don't see any currently configurable items. It also follows the [nvim-lua/kickstart](https://github.com/nvim-lua/kickstart.nvim#kickstartnvim) method of instruction as seen [here](https://github.com/nvim-lua/kickstart.nvim/blob/853b7c9d9e0e51cd7b9728242e309e33d6f77cac/init.lua#LL86C1-L87C42) and [here](https://github.com/nvim-lua/kickstart.nvim/blob/853b7c9d9e0e51cd7b9728242e309e33d6f77cac/init.lua#LL111C1-L111C41).

I think the comment should stay.  It will help guide those who don't know what it does and think they still need `setup()`.  One less question to answer in the future.

Does my explanation change your thoughts at all? Tell me and I will make the needed changes.

akinsho: > I don't know if config = true is any more performant

I don't think performance is a concern here in either case

Tbh I think the comment strays too far into teaching people how to use the `lazy.nvim` package manager which is absolutely not the domain of this plugin. If the key changes it's functionality or meaning I'll be chasing that plugin's setup around to make sure it stays up to date.

I think the balance here has to be pointing people in the right direction without putting ourselves into the position of doing it for them. They *must be able understand themselves what they are doing* whilst they might get confused that's fine they should go to the lazy docs what I don't want is for them to come here about anything lazy related which I believe will happen if we offer them too much info/the appearance that they should be looking here for all the answers and don't have to familiarise themselves with the lazy docs.

The vast majority of plugins do not have a comment like the above and I'd definitely err on the side of minimalism.

jidn: I made the modifications.

README.md
```diff
@@ -75,6 +75,7 @@ using `lazy.nvim`
         'nvim-lua/plenary.nvim',
         'stevearc/dressing.nvim', -- optional for vim.ui.select
     },
+    opts = true,
```
akinsho, line 78:
```suggestion
    config = true,
```

PR #271, opened 2023-07-07
fix: dap config selector ui, dap single request with no device

RyanCarrier:
dap launch config selector wasn't using the default ui option for this project

When selected a dap project that doesn't have deviceId specified, it will run twice, having to select the launch configuration twice.
Launch configuration is set prior to the first run, and is stored int he runner, but then cleaned up immediately after running


sidlatau:
Thanks for the PR, I adapted the code in https://github.com/nvim-flutter/flutter-tools.nvim/pull/424, it fixes the same issue.

RyanCarrier:
legend mate thank you

PR #272, opened 2023-07-09
fix(dap): define adapter and config when running standalone dart

### LINKED ISSUE #207: Setup dap for standalone dart apps
Hi there,
Maybe I am not understanding something correctly but when trying to debug Dart code (not flutter) the debugger complains because I didn't specify a device.

I switch between these two configs on for dart or flutter

Flutter:
```ts
      debugger = { 
          enabled = true,
          run_via_dap = true,

          register_configurations = function(_)
            require("dap").configurations.dart = {}
            require("dap.ext.vscode").load_launchjs()
          end
      },
```

Dart:
```ts
      debugger = { 
          enabled = true,
          run_via_dap = true,

          register_configurations = function(_)
              local dap = require('dap')
              dap.adapters.dart = {
                  type = "executable",
                  command = "dart",
                  args = { "debug_adapter" }
              }
             dap.configurations.dart = {
                 {
                      type = "dart",
                      request = "launch",
                      name = "Launch Dart Program",
                      program = "${file}",
                      cwd = "${workspaceFolder}",
                      args = { "--help" }, -- Note for Dart apps this is args, for Flutter apps toolArgs
                }
             }
            dap.load_launchjs()
          end
      },
```

If anyone has a solution for this I would appreciate.

igorlfs:
This should fix #207 

Couple questions: how can I test this? Should I use another pattern to detect if we're inside a flutter project?

Also, I'm getting an error when running the configuration (`DAP Error retrieving stack traces: getStack: (112) Service has disappeared`), but it seems to be caused by nvim-dap.

### DIFF
--- a/lua/flutter-tools/dap.lua
+++ b/lua/flutter-tools/dap.lua
@@ -1,3 +1,6 @@
+local lazy = require("flutter-tools.lazy")
+local path = lazy.require("flutter-tools.utils.path") ---@module "flutter-tools.utils.path"
+
 local ui = require("flutter-tools.ui")
 
 local success, dap = pcall(require, "dap")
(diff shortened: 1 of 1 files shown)


akinsho (approved):
LGTM 👍🏾 haven't tested it though so will have to see how it behaves for others

PR #285, opened 2023-09-13
feat: add install/uninstall cmd to menu

bruce3x:
I added two commands for installing and uninstalling the app, which is useful in the scenario of testing the initial installation of the app.

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -379,6 +379,58 @@ function M.fvm_use(sdk_name)
   end
 end
 
+---@type Job?
+local install_job = nil
+
+function M.install()
+  if not install_job then
+    executable.flutter(function(cmd)
+      local notify_timeout = 10000
+      install_job = Job:new({
+        command = cmd,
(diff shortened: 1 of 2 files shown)


akinsho:
@sidlatau any thoughts on this 🙏🏾 LGTM but not gonna be able to try it. Is it useful to have in your opinion

bruce3x:
This feature is inspired by the [adb-idea](https://github.com/pbreault/adb-idea) plugin.
image

sidlatau:
Looks like it does not work when the app uses flavors. For example, to uninstall it should construct such command respecting project configuration:
`flutter install --uninstall-only --flavor dev`
Flavors may be set up in .nvim.lua file in the project:
```lua
require("flutter-tools").setup_project({
	{
		name = "DEV",
		flavor = "dev"
	},
	{
		name = "PROD",
		flavor = "prod"
	},
})

```

sidlatau:
I will close it as it is implemented in /pull/315. Thanks @bruce3x for the initial code!

Issue #280, opened 2023-08-11
[BUG] flutter_tools_decorations.device is always nil


bruce3x:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After executing the `FlutterRun` command to start the flutter application, `vim.g.flutter_tools_decorations.device` is always `nil`.

### Expected Behavior

Its value should be the name of the currently running device.

### Steps To Reproduce

1. `:FlutterRun`
2. `:lua print(vim.g.flutter_tools_decorations.device)`

### Environment

```Markdown
- OS: macOS
- Flutter version: 3.10
- Is flutter in $PATH: yes
- neovim version:0.9.0
```

### Anything else?

I read the source code and tried to check the values of these two functions:
- [current_device()](https://github.com/akinsho/flutter-tools.nvim/blob/main/lua/flutter-tools/commands.lua#L41)
- [is_running()](https://github.com/akinsho/flutter-tools.nvim/blob/main/lua/flutter-tools/commands.lua#L43)

The result is :
- is_running = true
- current_device = nil

bruce3x:
Maybe broken by this commit
https://github.com/akinsho/flutter-tools.nvim/commit/6acf7d2ed2aeebf932808e99e64a4f71fe508606#diff-96c4b32009844045b7f1ce3515556a45d330b4b0c2567fa9fed85d0efb0a782fL111

PR #286, opened 2023-09-14
fix(commands): set current device while running project

bruce3x:
This PR fix https://github.com/akinsho/flutter-tools.nvim/issues/280

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -152,12 +152,22 @@ local function get_run_args(opts, conf)
   return args
 end
 
+--- @param args string[]
+--- @return Device?
+local function get_device_from_args(args)
+  for i = 1, #args - 1 do
+    if args[i] == "-d" then return { id = args[i + 1] } end
+  end
+end
+
 ---@param opts RunOpts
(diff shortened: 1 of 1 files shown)


bruce3x:
Now `flutter_tools_decorations.device` works very well with lualine plugin.

Here is my config
```lua
{
    "nvim-lualine/lualine.nvim",
    optional = true,
    opts = function(_, opts)
      local function is_available()
        return package.loaded["flutter-tools"] and require("flutter-tools.commands").is_running()
      end

      table.insert(opts.sections.lualine_x, 2, {
        function()
          if not is_available() then
            return
          end
          local device = vim.g.flutter_tools_decorations.device
          local device_name = device and (device.name or device.id) or "DEFAULT"
          return device_name .. " | " .. vim.g.flutter_tools_decorations.app_version
        end,
        color = "DiagnosticOk",
        cond = is_available,
      })
    end,
  }
```

image

bruce3x:
But when running `:FlutterRun` command directly on a single default device, I don't know how to get the device properties.

So I fallback to `"DEFAULT"` for lualine.

sidlatau (approved):

sidlatau:
Thanks, @bruce3x!

PR #290, opened 2023-09-18
fix: FlutterRename when files with import is opened in another buffer

### LINKED ISSUE #289: [BUG] FlutterRename command not updating imports
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The `FlutterRename` command doesn't  update import for another file if the other file is already opened in a different buffer

### Expected Behavior

`FlutterRename` renames the class, file name, and updates the import.

### Steps To Reproduce

1. Open two dart files in two different buffers, one with an import to the other. (ex. a `main.dart` and `MyWidget.dart`)
2. Use `:FlutterRename` on the Dart class (ex. `MyWidget` -> `MyNewWidget`)
3. The class and file name will change, but the import will not update. 
4. An error will be logged to stdout:

image

### Environment

```Markdown
- OS: macOS Ventura 13.5.2
- Flutter version: 3.10.6
- Is flutter in $PATH: true
- neovim version: 0.9.1
- LSP Settings:
  - renameFilesWithClasses = "always",
  - updateImportsOnRename = true,
```

### Anything else?

_No response_

m-salman-a:
Fixes #289

### DIFF
--- a/lua/flutter-tools/lsp/rename.lua
+++ b/lua/flutter-tools/lsp/rename.lua
@@ -89,8 +89,8 @@ function M.rename(new_name, options)
     params.newName = name
     local handler = client.handlers["textDocument/rename"] or lsp.handlers["textDocument/rename"]
     client.request("textDocument/rename", params, function(...)
-      handler(...)
       if result then lsp.util.apply_workspace_edit(result, client.offset_encoding) end
+      handler(...)
     end, bufnr)
   end
 


akinsho:
@m-salman-a I need a bit more context on what this PR is trying to do

sidlatau:
Hmm, I tested the original and changed version and I see no difference in behavior. As I understand it the reason why the import update is not applied is because the file had changes that were not saved yet. If I save changes - renaming works.

m-salman-a:
Hi, thanks for the quick reply. Upon recreating the issue, I found that you have to use the `FlutterRename` command more than once for the import to not be updated. I also made sure to save all the files before performing the second `FlutterRename`. Here is a video reproduction of the issue before and after applying the fix provided in the PR. [This is the minimal config that I used](https://github.com/m-salman-a/flutter-tools.nvim/blob/minimal_init/tests/minimal_init.lua).

## Before

https://github.com/akinsho/flutter-tools.nvim/assets/51153904/bfb8f8b3-2fb3-4431-ab54-27d3e1eaaebf

## After

https://github.com/akinsho/flutter-tools.nvim/assets/51153904/bd3ab94f-2d86-4f6f-bc81-9e4d1c90dfee

## Explanation

As for the explanation, the "Buffer .... newer than edits" error comes from this check [here](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/util.lua#L600) in the `vim.lsp.utils.apply_text_document_edit` function which is called from `vim.lsp.utils.apply_workspace_edit`. I'm currently not sure why this check is failing (if anyone can help that'd be great) but applying the workSpace/willRenameFiles request before calling the textDocument/rename handler makes this check pass.

sidlatau:
@m-salman-a thanks for the detailed explanation! Yea, when renaming more than once I am able to reproduce the issue, and your fix solves it. LGTM

sidlatau (approved):

PR #292, opened 2023-09-20
fix(dap): Flutter/Dart project detection improvement

### LINKED ISSUE #282: Flutter/Dart project detection improvement
hi, 

I ran into an issue that I couldn't start debugging. The error was about "dart:ui" inaccessible for this platform so I figured out it tries to run dart instead of flutter. For some reason I didn't have that `.metadata` file which is currently used to distinguish between pure dart and flutter projects. 
Maybe this logic could be improved? I'd be willing to contibute, I'm just a beginner in lua :)
  ```
    local is_flutter_project = vim.loop.fs_stat(path.join(root_dir, ".metadata"))
```

### PR DESCRIPTION
**Pull request by m-salman-a**
This PR aims to fix #282 by searching for the flutter sdk dependency from the `pubspec.yaml` file located in the root directory.  This is based on the fact that the `pubspec.yaml` in all flutter projects **requires** the flutter sdk dependency, as stated by [this specification](https://docs.flutter.dev/tools/pubspec). 

This is the simplest way that I can think of to differentiate between flutter or dart project, but any other suggestions are welcome 😀.

EDIT: This is still not a perfect solution as this still doesn't account for any comments that can be provided in the yaml file, but I think with this change and keeping the old behavior should be good enough for most users.

### DIFF
--- a/lua/flutter-tools/dap.lua
+++ b/lua/flutter-tools/dap.lua
@@ -17,8 +17,11 @@ function M.setup(config)
     local root_patterns = { ".git", "pubspec.yaml" }
     local current_dir = vim.fn.expand("%:p:h")
     local root_dir = path.find_root(root_patterns, current_dir) or current_dir
-    local pubspec_path = path.join(root_dir, "pubspec.yaml")
-    local is_flutter_project = pcall(vim.cmd, "silent 1vimgrep! /flutter:\\_s\\+sdk:\\_s\\+flutter\\C/j" .. pubspec_path)
+    local function search_pubspec_for_flutter_sdk()
+      local pubspec_path = path.join(root_dir, "pubspec.yaml")
+      return pcall(vim.cmd, "silent 1vimgrep! /sdk:\\_s\\+flutter\\C/j" .. pubspec_path)
+    end
+    local is_flutter_project = vim.loop.fs_stat(path.join(root_dir, ".metadata")) or search_pubspec_for_flutter_sdk()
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lua/flutter-tools/dap.lua [hunk 1]
```diff
@@ -17,7 +17,11 @@ function M.setup(config)
     local root_patterns = { ".git", "pubspec.yaml" }
     local current_dir = vim.fn.expand("%:p:h")
     local root_dir = path.find_root(root_patterns, current_dir) or current_dir
-    local is_flutter_project = vim.loop.fs_stat(path.join(root_dir, ".metadata"))
+    local function search_pubspec_for_flutter_sdk()
+      local pubspec_path = path.join(root_dir, "pubspec.yaml")
+      return pcall(vim.cmd, "silent 1vimgrep! /sdk:\\_s\\+flutter\\C/j" .. pubspec_path)
```
sidlatau, line 22: For the reference, I see Dart-Code uses a similar approach, it parses `yaml` and checks dependencies:
https://github.com/Dart-Code/Dart-Code/blob/0f3afe552c2cd918b0031fc1b77ede100b3fbea8/src/shared/utils/fs.ts#L163C11-L176

So I think this approach is valid, but I just do not fully understand regex: is underscore needed in '\\_s\\'?

sidlatau:
PR with almost the same logic was merged (https://github.com/akinsho/flutter-tools.nvim/pull/326) so closing this one.

PR #294, opened 2023-09-29
feat(dap): add custom commands to nvim-dap

ch-vik:
This simple PR aims to extend the available commands of `nvim-dap` by adding the custom requests `hotReload` and `hotRestart`. 

Being `customRequests` provided by the `flutter debug_adapter` and not part of the DAP standard payload, it makes sense that the flutter plugin extends these functionalities. 

This can also provide a means to use the hot reload functionality even in debug mode, where, as it is now, it is the only reliable way to run nested complex projects (with project configuration). 

- Extends the `DAP`commands repertoire with `hotReload` and `hotRestart` as `.hot-reload` and `.hot-restart`

### DIFF
--- a/lua/flutter-tools/dap.lua
+++ b/lua/flutter-tools/dap.lua
@@ -26,6 +26,13 @@ function M.setup(config)
         args = { "debug-adapter" },
       }
       opts.register_configurations(paths)
+      local repl = require("dap.repl")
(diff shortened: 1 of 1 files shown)


sidlatau (approved):
Tested, works, LGTM! :)

justjew:
Please, tell me how to use it

PR #295, opened 2023-09-30
fix(dap): fix cwd not being considered

ch-vik:
In `launch.json` it is possible to specify the current working directory (cwd). 
This can be used in complex projects to create configurations based on the main package, which in turn does not need the `program` param. 

It is more of a proposal than an actual fix; it heightens the inter-compatibility with how it functions in vs code. 

- Keeps `.cwd` if specified in the project config
- Changes `pick_one_sync` with `.pick_if_many` allowing the use of `vim.ui.select()`

### DIFF
--- a/lua/flutter-tools/runners/debugger_runner.lua
+++ b/lua/flutter-tools/runners/debugger_runner.lua
@@ -89,22 +89,24 @@ function DebuggerRunner:run(paths, args, cwd, on_run_data, on_run_exit)
   if launch_configuration_count == 0 then
     ui.notify("No launch configuration for DAP found", ui.ERROR)
     return
-  elseif launch_configuration_count == 1 then
-    launch_config = launch_configurations[1]
   else
-    launch_config = require("dap.ui").pick_one_sync(
+    require("dap.ui").pick_if_many(
       launch_configurations,
       "Select launch configuration",
-      function(item) return fmt("%s : %s", item.name, item.program, vim.inspect(item.args)) end
+      function(item)
+        return fmt("%s : %s | %s", item.name, item.program or item.cwd, vim.inspect(item.args))
+      end,
+      function(launch_config)
+        if not launch_config then return end
+        launch_config = vim.deepcopy(launch_config)
+        launch_config.cwd = launch_config.cwd or cwd
+        launch_config.args = vim.list_extend(launch_config.args or {}, args or {})
+        launch_config.dartSdkPath = paths.dart_sdk
+        launch_config.flutterSdkPath = paths.flutter_sdk
+        dap.run(launch_config)
+      end
     )
   end
-  if not launch_config then return end
-  launch_config = vim.deepcopy(launch_config)
-  launch_config.cwd = cwd
-  launch_config.args = vim.list_extend(launch_config.args or {}, args or {})
-  launch_config.dartSdkPath = paths.dart_sdk
-  launch_config.flutterSdkPath = paths.flutter_sdk
-  dap.run(launch_config)
 end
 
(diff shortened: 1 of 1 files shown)


sidlatau:
@ch-vik could you help me test this? What kind of `flutter-tools` setup you are using? I assume you have multiple configurations in `register_configurations`? 

Here is my setup: https://github.com/sidlatau/dotfiles/blob/master/nvim/.config/nvim/lua/plugins/flutter-tools.lua

ch-vik:
@sidlatau  Absolutely, I have actually tested it in our production project, but it can also be tested with a much simpler scenario (even if not realistic). 
First, create a multi-project/package folder, and from there, create a `.vscode/launch.json` file with the configurations for running each project, for example: 
```json
{
    "configurations": [
        {
            "name": "Flutter - debug",
            "request": "launch",
            "type": "dart",
            "cwd": "./app1/",
            "flutterMode": "debug"
        },
        {
            "name": "Flutter - profile",
            "request": "launch",
            "type": "dart",
            "cwd": "app1/",
            "flutterMode": "profile"
        }
        ,{
            "name": "Flutter 2 - debug",
            "request": "launch",
            "type": "dart",
            "cwd": "app2/",
            "flutterMode": "debug"
        }
    ]
}
```
Consider the project structure as this:
```
.
├── app1
│   ├── analysis_options.yaml
│   ├── android
│   ├── app1.iml
│   ├── build
│   ├── .dart_tool
│   ├── .gitignore
│   ├── .idea
│   ├── lib
│   ├── linux
│   ├── .metadata
│   ├── .packages
│   ├── pubspec.lock
│   ├── pubspec.yaml
│   ├── README.md
│   └── test
├── app2
│   ├── analysis_options.yaml
│   ├── android
│   ├── app2.iml
│   ├── build
│   ├── .dart_tool
│   ├── .gitignore
│   ├── .idea
│   ├── lib
│   ├── linux
│   ├── .metadata
│   ├── pubspec.lock
│   ├── pubspec.yaml
│   ├── README.md
│   ├── test
│   └── .vscode
└── .vscode
    └── launch.json
```
By configuring the `flutter-tools.nvim` package as this:
```lua
 {
    dir='~/nvim-projects/flutter-tools.nvim',
    lazy = false,
    dependencies = {
        'nvim-lua/plenary.nvim',
        'stevearc/dressing.nvim', -- optional for vim.ui.select
    },
    config = function ()
      require("flutter-tools").setup{
        debugger = {
          enabled = true,
          run_via_dap = true,
          register_configurations = function(_)
            require("dap").configurations.dart = {}
            require("dap.ext.vscode").load_launchjs()
          end,
        }
      }
    end,
},
```
You can run the project with one of the configurations specified in `launch.json`.

Now, if you run `:DapContinue` you can launch `app2` even if you are in a file from `app1` but you cannot do it with the `:FlutterRun` command because `cwd` is not considered.

With this PR, if the user specifies the `cwd` param, it is overwritten (as it is in `nvim-dap`) and also you can use `dressing.nvim` since it uses `vim.ui.select`.

sidlatau:
Thanks for the explanation!

I tested it. There may be some confusion when `require("flutter-tools").setup_project({...})` is used together with `require("dap.ext.vscode").load_launchjs()` - in that configuration now two selectors are shown for the user - one from project setup, another one for DAP launch configuration setup. But not sure if this is a problem, because they mostly duplicate each other. A workaround for this would be to not provide `register_configurations` in config:
```lua
 register_configurations = function(_)
            require("dap").configurations.dart = {}
            require("dap.ext.vscode").load_launchjs()
          end,
```

So LGTM!

sidlatau (approved):

Issue #298, opened 2023-10-07
[BUG] Insane delay after pressing enter


throwaway-squared:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

ive seen only one guy describe the same issue

when i press enter to get a new line in insert mode, it takes a good few seconds before nvim catches up and actually draws to the screen, although my input is still read and i can still type whats supposed to be in the next line, it just takes a while to actually show it

this only happens when flutter-tools is running (ive only tested it on the lib/main.dart file)

### Expected Behavior

upon pressing enter, a new line should be created without a delay of drawing it to the screen

### Steps To Reproduce

1. have my configuration (idk if this happens with other configs)
2. open a dart file (mainly lib/main.dart in a flutter project)
3. type some code
4. press enter
   should be noted that it seems to appear more often when splitting brackets in this manner]
   (| is the cursor): 
   1. appBar: AppBar(|)
   2. press enter

### Environment

Markdown
- OS:
```
Artix Linux
6.5.5-artix1-1
```
- Flutter version:
```
Flutter 3.13.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision ead455963c (11 days ago) • 2023-09-26 18:28:17 -0700
Engine • revision a794cf2681
Tools • Dart 3.1.3 • DevTools 2.25.0
```
- Is flutter in $PATH: yes definitely
- neovim: 
```
NVIM v0.9.2
Build type: Release
LuaJIT 2.1.1694285958
```
nvim config: https://codeberg.org/qube/dot-squared/src/branch/main/.config/nvim
```


### Anything else?

_No response_

gradunza5:
For whatever it's worth, I'm having this issue as well, and only with dartls. Other language servers I'm using (C#, lua, markdown, some others) don't show this issue. I haven't been able to find anything else on the internet about it.

koga25:
From what i searched, the problem is with [dart treesitter when it idents something](https://github.com/nvim-treesitter/nvim-treesitter/issues/4945). When i was searching about this i found two temporary solutions that worked for others:

[Removing dart treesitter](https://github.com/LunarVim/LunarVim/issues/4305#issuecomment-1681451926)
[Disabling dart treesitter ident](https://github.com/NvChad/NvChad/issues/2237#issuecomment-1653019941)

I don't know if this will help you, but it doesn't hurt to try.

arc41t3ct:
[...]

Issue #267, opened 2023-06-20
[BUG] Neovim Freezes 


Pratham16112002:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Current Behavior

When I am editing a dart file , press Enter i.e ( <CR> )  key my whole neovim gets halt for a second , this does not occur every time I press enter but most of the time . This is issue is only persist in the dartls 

https://github.com/akinsho/flutter-tools.nvim/assets/90999700/410f9ff5-10b2-4a96-b8c3-84f1aeb8f589

Also : After the neovim freezes and I can continuously press enter without any issue.  

### Expected Behavior

When I should press the enter key the neovim should never get halt , with .dart files 

https://github.com/akinsho/flutter-tools.nvim/assets/90999700/973a497a-d880-4ce4-8646-83c31e775885

In the above video enter key is working fine without any error , To be precise I am editing a pubspec.lock file 

### Steps To Reproduce

[issue_log.txt](https://github.com/akinsho/flutter-tools.nvim/files/11795926/issue_log.txt)
Attaching some logs that might make the issue more  clearer 

### Environment

```Markdown
- OS:Manjaro Linux x86_64 
- Flutter version: 3.10.3 
- Is flutter in $PATH: Yes
- neovim version: 0.9.0
```

### Anything else?

![image](https://github.com/akinsho/flutter-tools.nvim/assets/90999700/b924e1cd-093e-49f8-984b-32f91eb94be6)
LSP info SS 
#248 
This timeout issue looks similar to mine but I tried everything still nothing worked . 
`"server_request: callback result"	{  result = vim.NIL,  status = true}`  
Saw this log in  my log history .

nav-28:
I am having the same issue. I have tried a bare bones config and still get the error. Also, this only happens with dartls

akinsho:
> I am having the same issue. I have tried a bare bones config and still get the error. Also, this only happens with dartls

@nav-28 can you please confirm what a barebones config means in this case. Are you still using a configuration framework? can you please include precisely what minimal `init.lua` setup you are using that reliably reproduces this issue.

nav-28:
> > I am having the same issue. I have tried a bare bones config and still get the error. Also, this only happens with dartls
> 
> @nav-28 can you please confirm what a barebones config means in this case. Are you still using a configuration framework? can you please include precisely what minimal `init.lua` setup you are using that reliably reproduces this issue.

[...]

Issue #266, opened 2023-06-18
[BUG]  nvim gets halt 


Pratham16112002:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I press enter in insert Mode the neovim Gets halt for a second every time and then moves to second line and this is not happening with other lsp's just occur when I use this plugin 

### Expected Behavior

When I press enter in insert Mode the cursor should move to the next line immediately without halting 

### Steps To Reproduce

1. Linux 

### Environment

```Markdown
- OS:
- Flutter version:
- Is flutter in $PATH:
- neovim version:
```

akinsho:
@Pratham16112002 this isn't reproducible for me as you've described. This issue also contains next to no information other than saying you use some linux distro e.g. flutter version, neovim version etc. as such this is pretty impossible to debug or help with so will close this out without valid info.

PR #305, opened 2023-10-25
add warning for treesitter-dart and indentation

4e554c4c:
I thought a small addition to the readme might help reduce the number of issues (#298 #267 #266) created in this repo due to the related bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,9 @@ the excellent [dressing.nvim](https://github.com/stevearc/dressing.nvim).
 
 - You might encounter issues using this plugin on the `master` channel of flutter.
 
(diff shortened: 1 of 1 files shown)


akinsho:
@4e554c4c tbh I think a more general warning would useful to explain indentation and highlighting don't happen hear so no issues should be raised about that sort of thing here. I've almost certainly lost perspective as an almost exclusive neovim user, but I don't know why given the contents of the README already people thing that an LSP plugin would be the cause of any of this.

heddxh:
https://github.com/nvim-treesitter/nvim-treesitter/issues/4945
I think this warning is necessary.

sidlatau:
Closing this as upstream issue was closed.

PR #314, opened 2023-12-16
feat: add option to configure flutter mode via config

sidlatau:
It will add possibility to run app in `profile`, `release` modes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,10 @@ require('flutter-tools').setup_project({
     name = 'Web',
     device = 'chrome',
     flavor = 'WebApp'
+  },
+  {
+    name = 'Profile',
+    flutter_mode = 'profile', -- possible values: `debug`, `profile` or `release`, defaults to `debug`
   }
 })
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -320,6 +320,10 @@ require('flutter-tools').setup_project({
     name = 'Web',
     device = 'chrome',
     flavor = 'WebApp'
+  },
+  {
+    name = 'Profile',
+    flutter_mode = 'profile', -- possible values: `debug`, `profile` or `release`, defaults to `debug`
```
akinsho, line 326: Does this need to be prefixed with `flutter_` or can it just be mode or is it `flutter_` mode vs something like `dart_mode` (but not heard of that, but decidedly out of the loop atm)

sidlatau: Yea, naming is always difficult :)
I think it may be just `mode` or `build_mode`, as in docs: 'https://docs.flutter.dev/testing/build-modes'
`flutter_mode` was used to align with VSCode property `flutterMode` https://docs.flutter.dev/tools/vs-code#run-app-in-debug-profile-or-release-mode

PR #315, opened 2023-12-17
feat(commands): add install/uninstall commands to menu

sidlatau:
This is the updated version of akinsho/flutter-tools.nvim/pull/285, which additionally adds arguments from the project config.


akinsho (approved):
LGTM thanks @sidlatau 👍🏾

PR #319, opened 2023-12-25
feat: add performance_overlay, repaint_rainbow and slow_animations commands

sidlatau:
image


akinsho (approved):
LGTM but didn't do much testing

PR #326, opened 2024-01-21
add flutter dependency detection

mikeborodin:
### Flutter dependency detection
As mentioned in https://github.com/akinsho/flutter-tools.nvim/issues/282 the we need a better way to detect which configs to use - Dart or Flutter.   This PR intents to do that by matching the consequent lines in `pubspec.yaml`

#### Test cases I checked:

* Dart shelf example: `has flutter = false`
* Flutter counter example: `has flutter = true`
* Commented out 2 lines: `has flutter = false`
* Changed spacing (still valid yaml): `has flutter = true`
* Used tab instead of space: `has flutter = true`
* Removed lines: `has flutter = false`

### DIFF
--- a/lua/flutter-tools/dap.lua
+++ b/lua/flutter-tools/dap.lua
@@ -11,13 +11,20 @@ end
 
 local M = {}
 
+local function has_flutter_dependency_in_pubspec()
+  local pubspec = vim.fn.glob("pubspec.yaml")
+  if pubspec == "" then return false end
+  local pubspec_content = vim.fn.readfile(pubspec)
+  local joined_content = table.concat(pubspec_content, "\n")
+
+  local flutter_dependency = string.match(joined_content, "flutter:\n[%s\t]*sdk:[%s\t]*flutter")
+  return flutter_dependency ~= nil
+end
+
 function M.setup(config)
   local opts = config.debugger
   require("flutter-tools.executable").get(function(paths)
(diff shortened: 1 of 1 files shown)


akinsho:
@sidlatau thoughts? I haven't really worked with any non flutter dart projects

sidlatau:
> @sidlatau thoughts? I haven't really worked with any non flutter dart projects

LGTM. We already have PR with similar logic https://github.com/akinsho/flutter-tools.nvim/pull/292/files - but it is easier to understand regex here than in mentioned PR. 
Not sure which one is more performant and if that matters here.

mikeborodin:
> > @sidlatau thoughts? I haven't really worked with any non flutter dart projects
> 
> LGTM. We already have PR with similar logic https://github.com/akinsho/flutter-tools.nvim/pull/292/files - but it is easier to understand regex here than in mentioned PR. Not sure which one is more performant and if that matters here.

Yes I saw that one but seems like there was no progress in some time and in my approach I intentionally don't keep the previous behavior.

akinsho (approved):
LGTM 👍🏾

akinsho:
I imagine there might be some performance cost from using `fn.glob` and also `fn.readfile` as generally the `fn.*` methods do a trip through viml so are slower but I assume this function isn't called super often 🤷🏾‍♂️ we'll see

PR #333, opened 2024-02-23
feat(fvm): support FVM in monorepo setup

sidlatau:
As pointed in https://fvm.app/documentation/getting-started/faq#monorepo-support FVM could be set at the root of monorepo.


sidlatau:
I have run this for more than a month, no issues found, should be safe to use.

PR #335, opened 2024-03-15
Fix FlutterWidgetGuides highlight group overwriting custom colors

### LINKED ISSUE #325: [BUG] FlutterWidgetGuides highlight group gets overwritten on entering dart file
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If the highlight group `FlutterWidgetGuides` is set through `vim.api.nvim_set_hl()` in `(init.lua)`, it gets overwritten when opening a dart file to instead use the colors from the `Normal` group. 

### Expected Behavior

The `FlutterWidgetGuides` highlight group retains the custom colors set in `(init.lua).`

### Steps To Reproduce

N/A

### Environment

```Markdown
- OS: Arch linux
- Flutter version: 3.16.6-1
- Is flutter in $PATH: yes
- neovim version: v0.9.5
```

### Anything else?

I'm not sure if this is the correct way to override the highlight group but i was unable to find any alternative way, I've got it to work by setting up an autocmd on BuffEnter but it feels like a hacky way to do it.

If possible i would like to work on this issue by either adding a config parameter for the widget guide color or doing a check for the highlight group already being set. This is my first issue on GitHub, and I would appreciate guidance on the best approach.

wancup:
fix #325

I added a default option to set_hl to prevent overwriting custom colors in FlutterWidgetGuids.

### DIFF
--- a/lua/flutter-tools/guides.lua
+++ b/lua/flutter-tools/guides.lua
@@ -157,7 +157,9 @@ end
 
 function M.setup()
   local color = utils.get_hl("Normal", "fg")
-  if color and color ~= "" then utils.highlight(hl_group, { foreground = color }) end
+  if color and color ~= "" then
+    utils.highlight(hl_group, { foreground = color, default = true })
+  end
 end
 
 local function is_buf_valid(bufnum)


akinsho (approved):
Thanks 🙏🏾

PR #354, opened 2024-05-25
chore: if user set the path then not resolve

Decodetalkers:


Decodetalkers:
There are some project like rustup, itself will check the real executable, then run the executable, I have made such kind of project

https://github.com/Decodetalkers/flutterup

but if it find the final target called flutterup, the arg0 is `flutterup`, then it will not redirect to `flutter`, like what rustup does. and there are also another project do the same thing for flutter, but I have fogotten that project's name

lua/flutter-tools/executable.lua
```diff
@@ -127,7 +127,7 @@ function M.get(callback)
   end
 
   if config.flutter_path then
-    local flutter_path = fn.resolve(config.flutter_path)
+    local flutter_path = config.flutter_path
```
sidlatau, line 130: Hmm, this could be a breaking change. Would `config.flutter_lookup_cmd` work for you? If not, perhaps `config.flutter_path` could be a function. For example:
```lua
    local flutter_path = type(config.flutter_path) == "function" and config.flutter_path()
      or fn.resolve(config.flutter_path)

```

Decodetalkers: no. flutter_lookup_cmd won't work..

why here must be fn.resolve()? I think there is no need..

Decodetalkers: I see.. you need it to find dart and etc..

Decodetalkers: then what about add another config to set flutter sdk manually? I think this way to find flutter sdk is a little weird

sidlatau: We already have two configuration options, and introducing a third one might make things more confusing. These two options seem to cover most use cases. I agree that allowing `flutter_path` to be a function is a bit unconventional, but it serves as an escape hatch for manually providing the path, which I think is a reasonable compromise. Here's an example of how it would look:
```lua
{
  flutter_path = function()
    return 'path/to/flutter'
  end
}
```

Decodetalkers: But the thing that I am thinking about is that the way to find the flutter sdk is weird. Find the absolute path of flutter and  then find the path of the sdk.. but , yes, it will work for flutter. But does flutter have an official way to get the path of SDK?

Issue #353, opened 2024-05-24
[Feature] Pre-run callback in `setup_project{}`


JordanllHarper:
In my project, we have a script that needs to run before calling flutter run., given some flavor of the project.
It would be nice if we could create a "before run" function in the `"setup_project{}"` for this, or whatever naming works best for the project, that allows running some lua code before flutter run.

PR #358, opened 2024-06-20
Feature/353 pre run callback in setup project

JordanllHarper:
#353 -> Adds a pre run callback feature to setup_project, with some basic args exposed for users who want to use it without storing local config options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,8 @@ require('flutter-tools').setup_project({
       API_URL = 'https://dev.example.com/api',
       IS_DEV = true,
     },
+    pre_run_callback = nil, -- optional callback to run before the configuration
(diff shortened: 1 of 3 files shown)


sidlatau (approved):
LGTM

I personally did not have a need for such callback - just out of curiosity what use-case it helps to solve?

JordanllHarper:
On our project, we have a script that needs to run with the wanted flavor passed as an argument. This needs to happen before running the app.

JordanllHarper:
As an aside, do you want me to rename those commits to be conventional commit standard?

sidlatau:
> As an aside, do you want me to rename those commits to be conventional commit standard?

Yes, then we would have a nice changelog.

Issue #357, opened 2024-06-12
[BUG] "error drawing label" when cut a large block of code


justjew:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Sometimes when I cut (`d`) a big block of code, especially after `var qwer = ` until `;`, I am facing the following error.
```
error drawing label for /Users/path/to/my.dart on line 80. 
because:
```

If I disable `flutter-tools` and enable `dartls` using `lspconfig`, the error never appears.

Here's my `flutter-tools` config:
```lua
return {
  'akinsho/flutter-tools.nvim',
  lazy = false,
  dependencies = {
    'nvim-lua/plenary.nvim',
    'stevearc/dressing.nvim', -- optional for vim.ui.select
  },
  ft = 'dart',
  config = function()
    require('flutter-tools').setup {
      flutter_path = nil,
      flutter_lookup_cmd = 'asdf where flutter',
      statusline = {
        app_version = true,
      },
      lsp = {
        color = { -- show the derived colours for dart variables
          enabled = true, -- whether or not to highlight color variables at all, only supported on flutter >= 2.10
          background = false, -- highlight the background
          background_color = nil, -- required, when background is transparent (i.e. background_color = { r = 19, g = 17, b = 24},)
          foreground = true, -- highlight the foreground
          virtual_text = true, -- show the highlight using virtual text
          virtual_text_str = '■', -- the virtual text character to highlight
        },
      },
    }
    require('flutter-tools.lsp').attach()
  end,
}
```

### Expected Behavior

No error

### Steps To Reproduce

1. Cut a large block of some var definition from `=` to `;`
2. See error (sometimes)

### Environment

```Markdown
- OS: Mac OS Sonoma 14.2.1 (23C71)
- Flutter version: 3.22.2
- Is flutter in $PATH: of course
- neovim version: v0.9.5
```

JordanllHarper:
Not super helpful but have also seen this issue.

CrestNiraj12:
Any updates? This issue is so annoying.

akinsho:
For full visibility, this issue is going to need someone to get their hands dirty and dig in for a solution. As I've mentioned in other issues I'm not currently actively maintaining this and am fully dependent on *users* and *contributors* submitting PRs to fix issues like this.

JordanllHarper:
I'll be happy to have a look at this

JordanllHarper:
[...]

PR #360, opened 2024-06-25
fix(ui): render labels only in the document range

JordanllHarper:
Related issue: https://github.com/akinsho/flutter-tools.nvim/issues/357

Fix: Since the LSP lines can lag behind the document state, we only render labels that are in the buffer lines.


akinsho (approved):
LGTM 👍🏾

PR #363, opened 2024-06-25
feat: implement toggling the dev log

JordanllHarper:
This creates an autocommand for toggling the dev log to appear and disappear. Similar implementation to the outline window.

~~Also adjusts the default width for creating a new window, something was going wrong with the previous open win command regarding the width, when opening the toggle the second time. Tested and *seems* okay but might be worth a double check.~~


sidlatau:
https://github.com/nvim-flutter/flutter-tools.nvim/pull/411 was merged with almost the same functionality, but a bit more, like autoscrolling. Thanks, @JordanllHarper, for the pull request!

PR #371, opened 2024-08-13
feat: log_filter for flutter output

jonasbadstuebner:
I encountered spammy logs in flutter, that looked something like this:
```
[…]
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
E/GPUAUX  ( 4115): [AUX]GuiExtAuxCheckAuxPath:666: Null anb
I/gralloc4( 4115): @set_metadata: update dataspace from GM (0x00000000 -> 0x08010000)
[…]
```

When I execute `flutter run` via the CLI, I have a grep command that removes these lines for me. Sadly I did not find an easy way to use my wrapper function (with fvm) inside flutter-tools, so I decided to implement the feature directly in here.
It is the cleanest solution afterall, I suppose.

I also included an example of what the filter could look like. This is the one I use - because I only care about flutter logs on my dev phone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,9 @@ require("flutter-tools").setup {
   },
   dev_log = {
     enabled = true,
+    filter = nil, -- optional callback to filter the log
+    -- takes a log_line as string argument; returns a boolean or nil;
+    -- the log_line is only added to the output if the function returns true
     notify_errors = false, -- if there is an error whilst running then notify the user
     open_cmd = "tabedit", -- command to use to open the log buffer
   },
--- a/lua/flutter-tools/config.lua
+++ b/lua/flutter-tools/config.lua
@@ -128,6 +128,7 @@ local config = {
     __index = function(_, k) return k == "open_cmd" and get_split_cmd(0.3, 40) or nil end,
   }),
   dev_log = setmetatable({
+    filter = nil,
     enabled = true,
     notify_errors = false,
   }, {
--- a/lua/flutter-tools/log.lua
+++ b/lua/flutter-tools/log.lua
@@ -87,6 +87,7 @@ end
 function M.log(data, opts)
   if opts.enabled then
     if not exists() then create(opts) end
+    if opts.filter and not opts.filter(data) then return end
     append(M.buf, { data })
     autoscroll(M.buf, M.win)
   end


jonasbadstuebner:
Please let me know what you think!

README.md
```diff
@@ -189,6 +189,11 @@ problem so will conflict if both are set.
 
 ```lua
 
+-- if you want to use the regex log filter, I'd recommend defining it up here, so the
+-- RegEx does not have to be created again for every log line, but instead a single instance
+-- is reused.
+local filter_regex = vim.regex("^\\(D\\/\\|[VIWE]\\/flutter\\|.[^/]\\|\\s*$\\)")
```
sidlatau, line 195: I believe README should be "user friendly" and show the recommended plugin setup. I am not sure if we want to recommend using this regex for everyone.

README.md
```diff
@@ -238,6 +243,9 @@ require("flutter-tools").setup {
     prefix = ">", -- character to use for close tag e.g. > Widget
     enabled = true -- set to false to disable
   },
+  log_filter = function(log_line) -- you can filter the log here, if you encounter spammy logs
```
sidlatau, line 246: I think this function should belong to `dev_log` table/object (line 241).

lua/flutter-tools/commands.lua
```diff
@@ -67,7 +67,7 @@ end
 ---@param is_err boolean if this is stdout or stderr
 local function on_run_data(is_err, data)
   if is_err and config.dev_log.notify_errors then ui.notify(data, ui.ERROR, { timeout = 5000 }) end
-  dev_log.log(data, config.dev_log)
+  if config.log_filter == nil or config.log_filter(data) then dev_log.log(data, config.dev_log) end
```
sidlatau, line 70: It would be better to have filtering in the `dev_log.log` method. It has initialization logic, so for example if `log_filter` always returns `false`, `dev_log` buffer will never be shown even when triggered manually. That feels like a bug.

`config.log_filter == nil` may be replaced with `not config.log_filter`.

lua/flutter-tools/config.lua
```diff
@@ -127,6 +127,7 @@ local config = {
   }, {
     __index = function(_, k) return k == "open_cmd" and get_split_cmd(0.3, 40) or nil end,
   }),
+  -- log_filter = function(_) return true end,
```
sidlatau, line 130: The commented code does not add much value here.

sidlatau (commented):
The ability to filter logs looks useful 👍🏻. I added some comments about implementation.

lua/flutter-tools/config.lua
jonasbadstuebner: Hmm...I'm not sure, what is with config options that would not have to be set? In this case, simply checking, if a `log_filter` function is provided is faster (in terms of CPU load) than calling a function that always returns true. Since this is done for every log line, I figured, I should not include it in the default config, but it should be clear, that this is an option, if you need it.

lua/flutter-tools/commands.lua
jonasbadstuebner: You are absolutely right.

jonasbadstuebner: I would invert the logic and return early, if the log_filter is present, but returns false. This way, we don't have to wrap the whole "append to log" logic in a big if else statement.

README.md
jonasbadstuebner: I moved it, as I moved the log_filter into `dev_log.log`.
Will be updated in the next commit.

README.md
jonasbadstuebner: Hm...okay, so what would be a good approach? I would argue, that we can not really recommend a way to filter the log. Because the default behavior should be to see everything that is logged.

jonasbadstuebner:
Should I just call it `filter` if I put it into `dev_log` config table already?

jonasbadstuebner:
Should be clear, that this is a `dev_log`-`filter` then.

lua/flutter-tools/config.lua
sidlatau: We have an example here with `pre_run_callback` - I think we may be consistent there and do the same - in the config it could have `nil` - that may be enough to indicate the existence of this config option.

README.md
sidlatau: I think the best approach would be to remove this regex from README. We have a similar case with the `pre_run_callback` option:
 ```
    pre_run_callback = nil, -- optional callback to run before the configuration
    -- exposes a table containing name, target, flavour and device in the arguments
```
I think for consistency we may do the same - show it as `nil` but provide enough info about the structure of the function, what is the input and output.

sidlatau (approved):
LGTM, thanks for the PR!

PR #373, opened 2024-08-17
feat: Add option to set closing tag virtual text priority

devansh08:
This PR adds an option to set the priority for the closing tags virtual text.

This would help keep the order of virtual text items configurable, in cases when there are multiple virtual text items, eg.: git blame text from the `gitsigns` plugin or with LSP diagnostics text.

Example:

Current behavior (not preferable, especially in cases when commit messages are longer):
Screenshot_20240817_235218

When the priority is set (to 0, in this case) (preferable, as the tag is always kept next to the actual code):
Screenshot_20240817_235109

I've kept the default priority to be `0`, as I think logically one would want this text to be closest to the code. Open to discuss if there are any other thoughts on this.

Do let me know your thoughts on this. If this looks like a good change to have, I can add this in the docs (vimdoc and README) as well.
Thanks!

### DIFF
--- a/lua/flutter-tools/config.lua
+++ b/lua/flutter-tools/config.lua
@@ -109,6 +109,7 @@ local config = {
   closing_tags = {
     highlight = "Comment",
     prefix = "// ",
+    priority = 0,
     enabled = true,
   },
   lsp = {
--- a/lua/flutter-tools/config.lua
+++ b/lua/flutter-tools/config.lua
@@ -109,7 +109,7 @@ local config = {
   closing_tags = {
     highlight = "Comment",
     prefix = "// ",
-    priority = 0,
+    priority = 10,
     enabled = true,
   },
   lsp = {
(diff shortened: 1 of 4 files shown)


sidlatau:
@devansh08, thanks for the PR! The change looks good. Yes, documentation is needed, and it should be fine to merge.

devansh08:
Docs updated!

lua/flutter-tools/config.lua
```diff
@@ -109,6 +109,7 @@ local config = {
   closing_tags = {
     highlight = "Comment",
     prefix = "// ",
+    priority = 0,
```
sidlatau, line 112: @devansh08, the priority should align with what is mentioned in the documentation, as that represents the default configuration. I believe using the value 10, as stated in the documentation, would be better here. This would allow other plugins the flexibility to assign higher importance to their exmarks if needed.

devansh08: Ah, mentioning 10 in the docs was actually a mistake I just noticed :sweat_smile: But sure, what you say makes sense too. Better to keep it somewhat higher and avoid confusion since there might be a bunch of other plugins that do set it as 0, by default. Will update it.

devansh08: @sidlatau Updated!

sidlatau (approved):
LGTM, thanks for the PR!

PR #415, opened 2024-12-05
refactor: use plenary job to open devtools in browser

obemu:
My PR consists of 3 things:

1.  The `version.lua` module
     
     This module exposes the three functions `version`, `flutter_version` and `dart_version`. It also adds the classes `flutter.FlutterVersion` and `flutter.SemVer`. The functions can be called and used similarly to `get`, or `flutter` in the `executable.lua` module.
     Now this module is currently not being used anywhere else because the reason why I thought I needed it was a misunderstanding on my side, nevertheless I thought it could be useful in the future and kept it.
     
2. Improved types in the `executable.lua` module

    The type hint for the callback that must be provided to `get`, `flutter`, and `dart`, which receives the value to the flutter/dart binaries/SDKs was bad, because the function receives a `table<string, string>` and the user does not know what keys the table contains. I added the class `flutter.Paths` to fix that.
    
3. Improved `devtools.lua` module

    Bugfix in function `start_browser`: Using `vim.fn.jobstart` did not work on my machine (Debian 12, Linux 6.1.0-27-amd64 Debian 6.1.115-1 (2024-11-01) x86_64 GNU/Linux) so I replaced it with a `Job` from [plenary.nvim](https://github.com/nvim-lua/plenary.nvim).
    
    Bugfix in function `handle_start`: I noticed that the `json.params` table does not contain a `host` and `port` value the first time this function is called, so I added extra checks to ensure that the `host` and `port` are non-nil. I also decreased the duration of the notification duration from 20 seconds to 10, as this seems to be more than enough.
    
    I also made the function `start_browser` public, so that users of the plugin may call the function on their own.

### DIFF
--- a/lua/flutter-tools/dev_tools.lua
+++ b/lua/flutter-tools/dev_tools.lua
@@ -41,10 +41,13 @@ local function try_get_profiler_url_chrome(data)
   return data:match("(ws%:%/%/127%.0%.0%.1%:%d+/.+/ws)$")
 end
 
-local function start_browser()
+function M.start_browser()
   local auto_open_browser = config.dev_tools.auto_open_browser
   if not auto_open_browser then return end
+
   local url = M.get_profiler_url()
+  if not url then return end
+
   local open_command = utils.open_command()
   if not open_command then
     return ui.notify(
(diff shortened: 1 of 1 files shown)


sidlatau:
I see a lot of great work in this PR! However, I’m not entirely comfortable merging it as is since it introduces many changes in a single batch. Would it be possible to split it into smaller, more focused PRs? As a starting point, merging the start_browser fix would be a great step forward.

obemu:
I rebased the `feat_01` branch to only include the patch for the `devtools.lua` module, hopefully it can be merged now.

sidlatau (approved):
LGTM

Issue #434, opened 2025-01-30
[BUG] attempt to index local 'data' (a nil value)


JordanllHarper:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

On invoking a FlutterRun, if there is any error, I receive the following error:

![Image](https://github.com/user-attachments/assets/c4abfa62-fc25-4b07-91d5-d781d3a1a1a6)

### Expected Behavior

This error shouldn't occur. If anything goes wrong, it should exit gracefully (or error with a more descriptive error message).

### Steps To Reproduce

Prereq: Quit all emulators and connected devices.

Steps:
1. Create a new flutter project and cd into it
2. Open Neovim
3. Run FlutterRun. In the log, flutter should try to look for wireless devices.
4. An error should occur as described above.

### Environment

```Markdown
- OS: OSX
- Flutter version: 3.22.3
- Is flutter in $PATH: No - FVM
- neovim version: 0.11.0
```

JordanllHarper:
Looking into this bug myself, this looks to be an issue with the `M.handle_log` in `dev_tools.lua`; data from the job runner is giving nil values.

PR #436, opened 2025-02-06
fix: check for nil value when handle_log is called

JordanllHarper:
Addresses #434 . 

The job runner was returning nil values so added a check at the start of handle log to make sure data is not nil.


sidlatau (approved):

sidlatau:
@JordanllHarper, thanks for the PR!

PR #447, opened 2025-02-27
fix(outline): improve outline URI handling

sidlatau:
- Refactor outline URI storage to buffer level variable
- Add outline URI validation in get_outline_content checks
- Pass outline URI consistently through refresh_outline function


Ninja260:
@sidlatau Already tested the outline branch. It is much better. :100: I can't wait to see it on the main branch. I will close my PR as it is already solved with your commit.

PR #462, opened 2025-03-30
Check vim.fn.has explicitly for 1

belka-ew:
`vim.fn.has` returns a number, `0` or `1`, analogous to vimscript. But `0` isn't falsy in lua, but only `false` and `nil`. See https://github.com/nvim-flutter/flutter-tools.nvim/issues/461.

### DIFF
--- a/lua/flutter-tools/config.lua
+++ b/lua/flutter-tools/config.lua
@@ -36,7 +36,7 @@ local function validate_prefs(prefs)
       end
     )
   end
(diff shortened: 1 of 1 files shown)


sidlatau:
Thanks for the PR. More places needed the same fix, added it via https://github.com/nvim-flutter/flutter-tools.nvim/commit/6faf2c70bd56f1fe78620591a2bb73f4dc6f4870

PR #470, opened 2025-05-27
Fix flutter version check

obemu:
Prepend '--no-version-check' so that Flutter CLI tool does
not look for a new version.

Executing any command with the Flutter CLI tool makes the tool look
for a new version and optionally prints this banner:

```
┌─────────────────────────────────────────────────────────┐
│ A new version of Flutter is available!                  │
│                                                         │
│ To update to the latest version, run "flutter upgrade". │
└─────────────────────────────────────────────────────────┘
```

The problem is that this banner causes errors when parsing the
output of the running command. To fix this, we omit the version
check by default, but give the user the option to enable it.

### DIFF
--- a/lua/flutter-tools/commands.lua
+++ b/lua/flutter-tools/commands.lua
@@ -282,6 +282,25 @@ local function run(opts, project_conf, launch_config)
     local is_flutter_project = has_flutter_dependency_in_pubspec(cwd)
     if is_flutter_project then
       ui.notify("Starting flutter project...")
+
+      -- Prepend '--no-version-check' so that Flutter CLI tool does
+      -- not look for a new version.
+      --
+      -- Executing any command with the Flutter CLI tool makes the tool look
+      -- for a new version and optionally prints this banner:
+      --
+      -- ┌─────────────────────────────────────────────────────────┐
+      -- │ A new version of Flutter is available!                  │
+      -- │                                                         │
+      -- │ To update to the latest version, run "flutter upgrade". │
+      -- └─────────────────────────────────────────────────────────┘
+      --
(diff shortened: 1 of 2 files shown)


sidlatau:
Hi @obemu,
Thank you for this PR and for tackling the issue with the Flutter upgrade banner.
I've also seen this banner. Yes, it causes parsing error once, but I do find the update notification itself helpful.
I've been exploring a potentially more general solution for managing such CLI output in PR:
 - #471.
 
Could you please take a look at #471 when you have a chance and let me know if that approach might also work for you?
Thanks again for your contribution!

sidlatau:
> @sidlatau #471 seems like a good and future-proof (in case of any other banners/caveats) approach for this issue. However, I think it can be improved by having separate `global_run_options`, one for flutter and the other for dart projects, because the Flutter and Dart CLI tools do not have the same options. 

Yes, separating flutter and dart configs for `default_run_args` is more flexible solution, updated the PR.
So to achieve the same output as this PR this configuration may be used:
```lua
{
  default_run_args = { flutter = "--no-version-check" }
}
```


>I, too, think that the update notification is not entirely useless. However, If you really wish to keep it, then I reckon we must find a solution to tell the user via vim.notify that there is a new Flutter version instead of letting a command like FlutterRun complete with an error.

I created separate issue for this one: https://github.com/nvim-flutter/flutter-tools.nvim/issues/473

sidlatau:
Closing the PR, as we found I think more flexible solution there: https://github.com/nvim-flutter/flutter-tools.nvim/pull/471

Thanks again for helping to improve this plugin!

